PR #17, opened 2025-09-23
feat(server): CallAgentTool supports childThreadId to avoid thread conflicts

noa-lucent:
Addressed review comments:
- childThreadId is now required in the invocation schema.
- Improved description to clarify why this field exists (persistent conversation semantics) and how it computes the effective child thread id.
- Updated tests to always supply childThreadId.

PTAL.

### DIFF
--- a/apps/server/src/tools/call_agent.tool.ts
+++ b/apps/server/src/tools/call_agent.tool.ts
@@ -10,6 +10,13 @@ import { BaseMessage } from '@langchain/core/messages';
 const invocationSchema = z.object({
   input: z.string().min(1).describe('The message to forward to the target agent.'),
   context: z.any().optional().describe('Optional structured metadata; forwarded into TriggerMessage.info'),
+  childThreadId: z
+    .string()
+    .min(1)
+    .optional()
+    .describe(
+      'Optional child thread identifier. If provided, the target agent will be invoked in `${parentThreadId}__${childThreadId}`.',
+    ),
 });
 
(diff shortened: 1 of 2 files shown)


apps/server/src/tools/call_agent.tool.ts
```diff
@@ -10,6 +10,13 @@ import { BaseMessage } from '@langchain/core/messages';
 const invocationSchema = z.object({
   input: z.string().min(1).describe('The message to forward to the target agent.'),
   context: z.any().optional().describe('Optional structured metadata; forwarded into TriggerMessage.info'),
+  childThreadId: z
+    .string()
+    .min(1)
+    .optional()
```
vitramir: make it required

apps/server/src/tools/call_agent.tool.ts
```diff
@@ -10,6 +10,13 @@ import { BaseMessage } from '@langchain/core/messages';
...
+    .describe(
+      'Optional child thread identifier. If provided, the target agent will be invoked in `${parentThreadId}__${childThreadId}`.',
```
vitramir: Description is important for the consumer of the tool not for the engineers. Write clear reason for this field. I would say "it allows the caller to maintain multiple consistent conversations with child agents. conversation is persistent between invocations. if agent requires response or additional request the same thread id should be reused to continue conversation." but you can improve the wording to make it more precise.

vitramir (changes requested):

vitramir (approved):

PR #20, opened 2025-09-23
ci: add GitHub Actions workflow to run server tests

noa-lucent:
Renamed workflow to CI (ci.yml). Note: cannot manually dispatch because workflow_dispatch is not enabled. It will run automatically on push/PR. If you want me to add `workflow_dispatch` for manual triggers, I can push that small change.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
...
+  server-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
```
vitramir: return back cache pnpm. just setup pnpm before node

vitramir (changes requested):

vitramir (approved):

PR #21, opened 2025-09-23
chore: add ESLint (TS + import + prettier), keep no-explicit-any, and wire lint scripts

noa-lucent:
Addressed review: moved extensions into ESLint config and simplified scripts to `eslint .` with fix variant. Ready for another look.


package.json
```diff
@@ -4,7 +4,9 @@
   "description": "",
   "private": true,
   "scripts": {
-    "test": "vitest run"
+    "test": "vitest run",
+    "lint": "eslint . --ext .ts,.tsx",
+    "lint:fix": "eslint . --ext .ts,.tsx --fix"
```
vitramir: let's move extensions inside config

vitramir (changes requested):

PR #23, opened 2025-09-23
feat(server): template kind/title + rename bash→shell

noa-lucent:
Summary
- Enrich templates API with human-readable title and kind for each node (trigger | agent | tool)
- Rename BashCommandTool -> ShellTool (tool name: shell_command)

Changes
- apps/server/src/graph/types.ts
  - Add TemplateKind
  - Extend TemplateNodeSchema with title and kind
- apps/server/src/graph/templateRegistry.ts
  - register() now accepts meta: { title, kind }
  - toSchema() emits { name, title, kind, sourcePorts, targetPorts }
- apps/server/src/tools/shell_command.ts
  - New file replacing bash_command.ts
  - Class ShellTool, tool registered as 'shell_command'
- apps/server/src/templates.ts
  - Register titles/kinds for all templates
  - containerProvider titled "Workspace"
  - Register 'shellTool' instead of 'bashCommandTool'

API
- GET /api/templates response items now include title and kind fields in addition to ports

Notes
- No backward compatibility for bashCommandTool or bash_command per request.
- UI will use title/kind in a follow-up PR.

Testing
- Verify /api/templates includes fields: name, title, kind, sourcePorts, targetPorts
- Graph runtime unchanged aside from template identifiers.


apps/server/src/templates.ts
```diff
@@ -90,5 +92,6 @@ export function buildTemplateRegistry(deps: TemplateRegistryDeps): TemplateRegis
           containerProvider: { kind: 'method', create: 'setContainerProvider' },
         },
       },
+      { title: 'MCP Server', kind: 'tool' }
```
vitramir: MCP server should be of kind MCP

apps/server/src/tools/shell_command.ts
vitramir, line 1: new shell_command file was created but old bash tool wasn't removed

vitramir (changes requested):

noa-lucent (commented):
Addressed:
- Set MCP Server kind to 'mcp' and extended TemplateKind accordingly.
- Confirmed old bash tool file is removed; only shell_command.ts remains and references updated.

Let me know if you’d prefer MCP to remain under 'tool' instead of its own kind; easy to switch back.

noa-lucent (commented):
Removed legacy file apps/server/src/tools/bash_command.ts in the feature branch. Thanks for catching that!

PR #36, opened 2025-09-24
feat(container): platform support for image pull and container creation (typed, no-any)

noa-lucent:
Addressed review: removed `any` casts; added module augmentation for dockerode types.

Changes since opening:
- Introduced apps/server/src/types/dockerode-augmentation.d.ts with:
  - ContainerCreateOptions.Platform and Entrypoint typings
  - ImagePullOptions.platform typing
- ContainerService:
  - ensureImage now uses typed ImagePullOptions and `docker.pull(image, pullOpts, cb)` when platform provided
  - create options set Platform and Entrypoint directly (no casts)
  - Stream handling uses typed PassThrough/Writable instead of `as any` sink objects
- No behavioral changes; only typing and minor refactor to satisfy strict TS.

This ensures all passed values comply with interfaces under `strict` tsconfig.

### DIFF
--- a/apps/server/src/services/container.service.ts
+++ b/apps/server/src/services/container.service.ts
@@ -1,4 +1,5 @@
-import Docker, { ContainerCreateOptions, Exec } from 'dockerode';
+import Docker, { ContainerCreateOptions, Exec, ImagePullOptions } from 'dockerode';
+import { PassThrough, Writable } from 'node:stream';
 import { ContainerEntity } from '../entities/container.entity';
 import { LoggerService } from './logger.service';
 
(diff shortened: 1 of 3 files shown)


apps/server/src/services/container.service.ts
```diff
@@ -73,7 +74,13 @@ export class ContainerService {
             }
           },
         );
-      });
+      };
+      if (platform) {
+        // Pass platform to ensure correct architecture is pulled; use any to avoid type issues across dockerode versions
+        (this.docker as any).pull(image, { platform }, cb);
```
vitramir: do not use any

apps/server/src/services/container.service.ts
```diff
@@ -105,7 +112,12 @@ export class ContainerService {
       AttachStdout: true,
       AttachStderr: true,
       Labels: optsWithDefaults.labels,
-    };
+    } as any;
+
+    if (optsWithDefaults.platform) {
+      // Prefer platform at container creation time if provided
+      (createOptions as any).Platform = optsWithDefaults.platform;
```
vitramir: do not use any

vitramir (changes requested):

Issue #66, opened 2025-10-02
Dynamic Configuration shows "Unsupported field schema ... undefined" after MCP server restart until page reload


noa-lucent:
Summary
After stopping and starting an MCP server, the Agents UI Dynamic Configuration panel briefly shows an error and an incomplete state until the page is reloaded.

Observed behavior
Dynamic Configuration
Unsupported field schema for field root: Unknown field type undefined.
{
  "ready": false
}
If the page is reloaded, the proper configuration is then loaded.

Expected behavior
Dynamic Configuration should load the correct schema and values automatically after the MCP server is restarted, without requiring a full page reload.

Suspected cause
Likely a race condition or missing re-fetch of the dynamic configuration schema/state upon MCP server reconnect/start. UI may render before the schema is available, causing schema to be undefined. Page reload resets state and triggers correct fetch.

Steps to reproduce
1) Start the Agents app
2) Stop the MCP server for an agent
3) Start the MCP server again
4) Navigate to (or observe) the Dynamic Configuration panel
5) Notice the error message above; a reload resolves it

Requested
- Reproduce and identify root cause
- Implement a fix to ensure schema/state is refreshed on reconnect/start
- Add tests (unit/e2e) to prevent regression
- Consider graceful handling (loading state) when schema is temporarily unavailable

Additional info
Reporter will provide browser/environment details and any console logs in comments.

noa-lucent:
Thanks for the report. We’ve completed an initial investigation and identified the likely root cause and a minimal fix plan.

Root cause (summary)
- The UI Dynamic Configuration form uses @rjsf/core and crashes with “Unsupported field schema… undefined” when it receives an invalid schema (e.g., an empty object or a wrapper like { ready: false }).
- The schema fetcher (apps/ui/src/lib/graph/api.ts:getDynamicConfigSchema) returns the whole response (including { ready:false }) or {} when the server isn’t ready, which is then passed to RJSF.
- React Query caches that result for ~60s. When the MCP server becomes ready later, the UI may render using the cached invalid schema, causing the error until a full page reload clears/refreshes cache.
- Additionally, the backend does not emit a follow-up `node_status` event when the MCP server actually finishes discovery and becomes ready, so the UI may not know to refetch.

Fix plan
UI (apps/ui):
1) Harden getDynamicConfigSchema to only return a valid JSON Schema object. If the response is { ready:false } or lacks a valid schema root (type/properties/$ref), return null.
[...]

PR #67, opened 2025-10-02
Server: emit node_status on ready after MCP restart (#66)

noa-lucent:
Server-side enhancement for Issue #66: ensure a follow-up node_status is broadcast when an MCP node becomes ready after discovery.

What changed
- New ReadinessWatcher helper that polls runtime.getNodeStatus(nodeId) ~500ms until either provisionStatus.state === "ready" OR dynamicConfigReady === true, or timeout (~30s). On readiness it calls existing emitStatus(nodeId) to broadcast the latest status via socket.
- Debounce: Map-based guard to avoid multiple concurrent watchers per node; coalesces starts. Cleaned up on deprovision and server shutdown.
- Hooked into POST /graph/nodes/:nodeId/actions: start watcher on provision; stop on deprovision. Initial status still emitted immediately (existing behavior).

Tests
- Added unit tests for watcher behavior: emits once on ready; no emit on timeout; cancels when node disappears. Extracted watcher into utils for testability.

Why
- After LocalMCPServer finishes discovery, dynamicConfigReady can flip to true later. The UI caches node status and schema, so this follow-up broadcast prompts clients to re-fetch the dynamic config schema without a page reload.

Notes
- Minimal change footprint; no refactor of liveGraph.manager. Timeout and intervals are conservative, easy to tune.
- Safeguards: debounce, timeout, and stop on deprovision to prevent spam.

### DIFF
--- a/apps/server/__tests__/readinessWatcher.test.ts
+++ b/apps/server/__tests__/readinessWatcher.test.ts
@@ -48,17 +48,15 @@ describe('ReadinessWatcher', () => {
       { provisionStatus: { state: 'ready' }, dynamicConfigReady: false },
     ]);
 
+    // Start twice immediately to verify debounce of concurrent watchers
     watcher.start(nodeId);
+    watcher.start(nodeId);
+
     // wait enough time for polling
     await new Promise((r) => setTimeout(r, 50));
 
     expect(emit).toHaveBeenCalledTimes(1);
     expect(emit).toHaveBeenCalledWith(nodeId);
-
-    // ensure subsequent starts are debounced while active
-    watcher.start(nodeId);
-    await new Promise((r) => setTimeout(r, 20));
-    expect(emit).toHaveBeenCalledTimes(1);
 
     watcher.stopAll();
   });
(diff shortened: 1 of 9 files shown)


noa-lucent (commented):
UI-side fix for Issue #66 pushed to this branch.

Proposed PR title: "UI: fix Dynamic Config schema race after MCP restart (#66)"

What changed (UI):
- apps/ui/src/lib/graph/api.ts: Normalize getDynamicConfigSchema to return a valid JSON Schema object or null; updated return type to Promise<Record<string, unknown> | null>. Added minimal schema shape guard (type/properties/$ref).
- apps/ui/src/components/graph/DynamicConfigForm.tsx: Guard against invalid schema. When dynamicConfigReady===true but schema is null/invalid, render a small "Loading dynamic config..." placeholder and trigger a one-time refetch. Avoid passing {} to RJSF.
- apps/ui/src/lib/graph/hooks.ts: On node_status with dynamicConfigReady===true for the active node, proactively invalidate the schema query ["graph","node",nodeId,"dynamic","schema"] with a short debounce to refetch immediately.
- Tests: Updated api.test.ts to cover normalization; extended DynamicConfigForm.test.tsx to ensure no crash with null/invalid schema and refetch on ready.

Why: The UI previously cached a wrapper/empty response from the dynamic config endpoint during MCP startup, causing RJSF to throw "Unsupported field schema ... type undefined" until a full page reload. These changes prevent invalid schemas from reaching the form and ensure we refetch as soon as the server signals readiness.

Acceptance:
- No RJSF error after stopping/starting MCP without reloading; Dynamic Configuration loads automatically once ready.
- No regressions in existing forms.

If you prefer, we can split server and UI pieces into separate PRs; given the branch already exists for #66, I added the UI work here. Please also consider retitling this PR accordingly or let me know and I will open a follow-up PR.

PR #71, opened 2025-10-03
feat(server): add workspace platform support (linux/amd64, linux/arm64) and enforce non-reuse on mismatch

noa-lucent:
This PR adds optional platform support for the Workspace container provider with strict values and enforces non‑reuse when the platform mismatches.

What changed
- ContainerProviderStaticConfigSchema gains `platform?: z.enum(['linux/amd64','linux/arm64'])`.
- ContainerProviderEntity stores `platform` in its config and forwards it to ContainerService.start().
- Enforce non-reuse: if a container is found by labels but the requested platform is defined and differs from the container’s label `hautech.ai/platform` (or is missing), the old container is stopped and removed and a new one is created.
- ContainerService:
  - ContainerOpts now includes `platform?: 'linux/amd64' | 'linux/arm64'`.
  - ensureImage(image, platform) passes platform to docker.pull.
  - start() threads platform to ensureImage and docker.createContainer as a query param, and adds label `hautech.ai/platform` to Labels when defined.
- Tests:
  - apps/server/__tests__/container.service.platform.test.ts mocks dockerode and verifies pull/createContainer args and labels for platform set/unset.
  - apps/server/__tests__/containerProvider.entity.platform.test.ts covers reuse policy for mismatch and missing label, reuse when undefined, and schema rejects invalid values.
- Docs: README and technical overview updated to explain the new platform field, allowed values, behavior, and caveats.

Acceptance
- When platform is set, docker.pull includes { platform }, docker.createContainer gets platform as a query param, new containers are labeled `hautech.ai/platform`, and reuse is prevented on mismatch/missing label.
- When platform is undefined, behavior remains unchanged; no label added and reuse works as before.
- All unit tests pass without requiring Docker.

Notes
- dockerode maps unknown top-level options to query params; types casted where needed.
- Changes are scoped to apps/server only.


apps/server/src/services/container.service.ts
noa-lucent, line 1: Great addition overall. A few type-safety nits and a minor robustness tweak:

1) Avoid `as any` in ensureImage() pull call
- Current:
  if (platform) {
    (this.docker as any).pull(image, { platform }, cb as any);
  } else {
    this.docker.pull(image, cb as any);
  }
- Suggestion (no `any`):
  const callback: (err: unknown, stream?: NodeJS.ReadableStream) => void = cb;
  if (platform) {
    this.docker.pull(
      image,
      { platform } as unknown as Parameters<Docker['pull']>[1],
      callback as Parameters<Docker['pull']>[2],
    );
  } else {
    this.docker.pull(image, callback as Parameters<Docker['pull']>[1]);
  }
This keeps the runtime behavior while removing `any` casts.

Also, you can tighten the progress event type a bit:
- Replace `(event: any)` with something like `(event: { status?: string; id?: string })`.

2) Avoid `as any` in createContainer
- Current: `const container = await (this.docker as any).createContainer(createOptions as any);`
- Suggestion:
  type CreateOptsWithPlatform = ContainerCreateOptions & { name?: string; platform?: string };
  const createOptions: CreateOptsWithPlatform = { ... };
  const container = await this.docker.createContainer(createOptions as unknown as ContainerCreateOptions);
No `any` required and still threads the `platform` query param.

3) getContainerLabels: remove `as any`
- Current: `return (details?.Config as any)?.Labels as Record<string, string> | undefined;`
- Suggestion: Dockerode types expose `details.Config?.Labels` already. Prefer:
  `return details.Config?.Labels ?? undefined;`
If Typescript complains, cast via `as unknown as Record<string, string> | undefined`, but avoid `any`.

4) Minor: removeContainer indentation
There’s a small indentation hiccup inside removeContainer after this change. Not functionally relevant, but please fix to match the file’s formatting conventions.

apps/server/src/entities/containerProvider.entity.ts
noa-lucent, line 1: Nit: avoid `as any` casting. Here: `const requestedPlatform = this.cfg?.platform ?? (this.opts as any)?.platform;` — `opts` is already typed as `ContainerOpts` (which now includes `platform`), so this can be `const requestedPlatform = this.cfg?.platform ?? this.opts.platform;` with no cast.

noa-lucent, line 1: Nice reuse policy addition. Two improvements to align with our style rules and reduce duplication:

1) Avoid `as any` for opts.platform
- Current: `const requestedPlatform = this.cfg?.platform ?? (this.opts as any)?.platform;`
- Since `this.opts` is typed as `ContainerOpts`, you can just use:
  `const requestedPlatform = this.cfg?.platform ?? this.opts.platform;`
No cast needed.

2) DRY the platform label key
- The string literal `'hautech.ai/platform'` is used here (for comparison) and in ContainerService (for labeling). Suggest extracting a shared constant, e.g.:
  `export const PLATFORM_LABEL = 'hautech.ai/platform';`
  in a small module like `apps/server/src/constants.ts` and importing it in both places. This prevents typos and eases future changes.

apps/server/__tests__/container.service.platform.test.ts
noa-lucent, line 1: Tests look solid and CI-safe. Two small suggestions:

- Prefer avoiding `any` in the MockDocker:
  - `modem.followProgress: vi.fn((stream: NodeJS.ReadableStream, cb: (err?: Error) => void) => { ... })`
  - `pull = vi.fn((image: string, optsOrCb?: object | ((err?: Error, stream?: NodeJS.ReadableStream) => void), maybeCb?: (err?: Error, stream?: NodeJS.ReadableStream) => void) => { ... })`
  This maintains type-safety without changing behavior.

- In the first test, you already assert `createOpts.platform` and the label. It may also be worth asserting that `createOpts.Labels` still contains pre-existing labels when provided (e.g., pass one via opts and expect it to survive), to protect against clobbering regressions.

apps/server/__tests__/containerProvider.entity.platform.test.ts
noa-lucent, line 1: Great coverage of the reuse paths and schema validation. Two nits:

- Minor typing cleanups to avoid `as any`:
  - When building the `svc` mock, prefer explicit types for function properties instead of `as any` (e.g., `findContainerByLabels: vi.fn(async () => undefined) as unknown as ContainerService['findContainerByLabels']`).
  - Same for `start` and `getContainerLabels`.

- Consider adding one more assertion to ensure that when platform mismatches, the `labels` used to find containers still include the thread label (guard against future refactors that might break concurrency semantics). A simple spy on `findContainerByLabels` calls should suffice.

apps/server/src/services/container.service.ts
noa-lucent, line 1: Type-safety: please remove the `as any` casts around docker.pull. Dockerode’s typings support passing an options object to pull. You can do:

await new Promise<void>((resolve, reject) => {
  const cb = (err: Error | undefined, stream: NodeJS.ReadableStream | undefined) => { /* unchanged */ };
  this.docker.pull(image, platform ? { platform } : undefined, cb);
});

This avoids both `(this.docker as any).pull` and `cb as any` while keeping the same behavior.

noa-lucent, line 1: Bug/Nit: indentation lost here and semantic change. The method previously did an inspect first and then remove. Now it logs and calls remove directly, but the inspect line was repurposed to implement getContainerLabels below. The dedent also breaks our Prettier rule.

Suggest restoring:
- Keep removeContainer as before (just getContainer+remove), but fix indentation.
- Implement getContainerLabels separately without altering removeContainer.

Concretely:
async removeContainer(containerId: string, force = false): Promise<void> {
  this.logger.info(`Removing container cid=${containerId.substring(0, 12)} force=${force}`);
  const container = this.docker.getContainer(containerId);
  await container.remove({ force });
}

…then keep getContainerLabels as added.

noa-lucent, line 1: Minimize casts on createContainer too. We can use a typed helper for the query param without forcing `(this.docker as any)`:

const createOptions: ContainerCreateOptions & { name?: string } = {/* as defined, but no platform key here */};
const query: { platform?: 'linux/amd64' | 'linux/arm64' } = optsWithDefaults.platform ? { platform: optsWithDefaults.platform } : {};
const container = await this.docker.createContainer({ ...createOptions, ...query } as any);

Alternatively, narrow the type once with an interface that extends ContainerCreateOptions to include `platform?: string` so you don’t need `as any` at call sites.

noa-lucent, line 1: Label merge looks good. One suggestion: guard against Labels being undefined in inspect results when reading them elsewhere, but here at creation time you already spread from `{}` so it’s fine.

apps/server/src/entities/containerProvider.entity.ts
noa-lucent, line 1: Reuse enforcement: thanks for awaiting stop() then remove(true). Minor: handle stop errors gracefully (e.g., already stopped) like ContainerService.stopContainer does. Consider wrapping stop in try/catch and still calling remove(true) to avoid orphans.

apps/server/__tests__/container.service.platform.test.ts
noa-lucent, line 1: Test fragility: this expectation assumes docker.pull signature without opts when platform undefined. Dockerode accepts (image, opts?, cb). Our code calls pull(image, cb) when platform undefined, so this passes, but to be safer assert that args[1] is either a function or an options object without platform:

if (typeof pullArgs[1] === 'function') { /* ok */ } else { expect(pullArgs[1]?.platform).toBeUndefined(); }

apps/server/src/services/container.service.ts
```diff
@@ -105,13 +112,16 @@ export class ContainerService {
       Tty: optsWithDefaults.tty ?? false,
       AttachStdout: true,
       AttachStderr: true,
-      Labels: optsWithDefaults.labels,
+      Labels: {
+        ...(optsWithDefaults.labels || {}),
```
noa-lucent, line 118: Nit: consider extracting the platform label key into a shared constant (e.g., PLATFORM_LABEL = 'hautech.ai/platform') and reusing it here and in ContainerProviderEntity to avoid string duplication/typos.

docs/README.md
noa-lucent, line 1: Docs: nice addition. Suggest adding a brief note that Docker Engine server-side must support the requested platform (qemu/binfmt_misc for emulation) and that not all registries publish multi-arch images with the same tag.

apps/server/__tests__/containerProvider.entity.platform.test.ts
noa-lucent, line 1: Tests: can we also assert that when platform is undefined and we reuse, start() is not called and remove() is not called (you already assert stop not called). You did assert start not called, good. Consider also asserting remove not called for completeness.

noa-lucent (commented):
Left several inline comments. Once those small type-safety and robustness tweaks are addressed (primarily removing `any` casts, using a shared constant for the platform label, and catching NotFound on removal), this looks ready to ship. Nice work on tests and docs.

noa-lucent (commented):
Thanks for the well-scoped PR — functionality and tests look solid overall. I verified each requested area; a few type-safety and small robustness issues remain before merge.

Summary by scope
1) Schema & UI exposure
- ContainerProviderStaticConfigSchema includes platform?: z.enum(['linux/amd64','linux/arm64']).
- apps/server/src/templates.ts passes the zod schema via toJSONSchema, so the enum will render as a select dropdown in the Workspace node. No regression found.

2) Service changes
- ContainerOpts adds platform?: 'linux/amd64' | 'linux/arm64' — OK.
- ensureImage(image, platform) threads platform to docker.pull — OK functionally, but please remove casts and use the typed overload for pull (see inline comment with example).
- start(opts) threads platform to ensureImage — OK.
- docker.createContainer: passing platform as a query param is correct, but please avoid the `as any` casts. Prefer narrowing with unknown and a localized helper type or extending the options type so the cast is centralized and safe. Also only include the `platform` key when defined to avoid `?platform=undefined`.
- Labels: 'hautech.ai/platform' is added only when platform is defined; merge preserves existing labels — OK.

3) Reuse logic enforcement
- In provide(threadId): logic to stop/remove on platform mismatch or missing label is correct. Please handle stop() errors gracefully (e.g., already stopped) and still call remove(true) to avoid orphans. Also remove the unnecessary `(this.opts as any)` cast; opts is already typed.

4) Tests (CI-safe, mocked)
- container.service.platform.test.ts validates pull/create and label behaviors with and without platform — OK. One minor fragility around the pull signature when platform is undefined; I suggested a safer assertion form.
- containerProvider.entity.platform.test.ts covers all reuse cases and schema validation — OK. Consider adding an explicit assertion that remove() is not called when platform is undefined and we reuse (completeness).

5) Docs
- README and technical-overview updated; no references to windows/* found. Consider adding a brief note about Docker Engine support for cross-arch via qemu/binfmt_misc and that not all tags are multi-arch.

6) Nits and correctness
- Remove `as any` usage in src:
  - container.service.ts: docker.pull invocation and createContainer call; prefer typed overloads and localized unknown → specific function type assertions.
  - container.service.ts: getContainerLabels can rely on ContainerInspectInfo types and avoid casts.
  - containerProvider.entity.ts: `(this.opts as any)?.platform` not needed; use `this.opts.platform`.
- Indentation: fix the removeContainer method’s dedent (Prettier violation) and keep it functionally scoped to removal; do not mix in logic for getContainerLabels.
- Only include the `platform` key in the create options when defined to avoid accidental query pollution.

Once those are addressed, I’m happy to re-review and approve.

docs/README.md
noa-lucent, line 1: Thanks for the feature! High-level LGTM. I’m doing a focused pass per the review checklist the team provided. I’ll leave targeted line comments and then summarize with an overall verdict.

apps/server/src/entities/containerProvider.entity.ts
```diff
@@ -44,13 +48,43 @@ export class ContainerProviderEntity {
   async provide(threadId: string) {
     const labels = this.idLabels(threadId);
     let container: ContainerEntity | undefined = await this.containerService.findContainerByLabels(labels);
+
+    // Enforce non-reuse on platform mismatch if a platform is requested now
+    const requestedPlatform = this.cfg?.platform ?? this.opts.platform;
```
noa-lucent, line 56: Nit: avoid casting to any. opts is already typed as ContainerOpts (which includes platform), so this can be:

const requestedPlatform = this.cfg?.platform ?? this.opts.platform;

apps/server/src/services/container.service.ts
```diff
@@ -55,25 +57,27 @@ export class ContainerService {
...
             if (doneErr) return reject(doneErr);
             this.logger.info(`Finished pulling image '${image}'`);
             resolve();
           },
-          (event: any) => {
+          (event: { status?: string; id?: string }) => {
             if (event?.status && event?.id) {
               this.logger.debug(`${event.id}: ${event.status}`);
             } else if (event?.status) {
               this.logger.debug(event.status);
```
noa-lucent, line 76: Type-safety: please avoid `(this.docker as any).pull` and `cb as any`. We can add a local module augmentation for dockerode to type the platform option and callback correctly, e.g. in `apps/server/src/types/dockerode.d.ts`:

```ts
// types/dockerode.d.ts
import 'dockerode';
declare module 'dockerode' {
  interface ImagePullOptions { platform?: string }
}
```

Then call:

```ts
docker.pull(image, platform ? { platform } : undefined, cb);
```

No `any` casts needed.

apps/server/src/services/container.service.ts
```diff
@@ -105,13 +111,18 @@ export class ContainerService {
...
       AttachStdout: true,
       AttachStderr: true,
-      Labels: optsWithDefaults.labels,
+      Labels: {
+        ...(optsWithDefaults.labels || {}),
+        ...(optsWithDefaults.platform ? { [PLATFORM_LABEL]: optsWithDefaults.platform } : {}),
+      },
     };
 
+    const query = optsWithDefaults.platform ? { platform: optsWithDefaults.platform } : undefined;
```
noa-lucent: Same here: please avoid `(this.docker as any).createContainer(createOptions as any)`. Consider a local type augmentation so `ContainerCreateOptions` allows an optional `platform?: string` (and `name?: string` if you want to keep it colocated) to avoid casts. Example augmentation in `src/types/dockerode.d.ts`:

```ts
import 'dockerode';
declare module 'dockerode' {
  interface ContainerCreateOptions {
    // Query-string keys allowed by Docker API that dockerode forwards
    name?: string;
    platform?: string;
  }
}
```

Then call `docker.createContainer(createOptions)` without casts.

apps/server/src/services/container.service.ts
```diff
@@ -105,13 +111,18 @@ export class ContainerService {
       Tty: optsWithDefaults.tty ?? false,
       AttachStdout: true,
```
noa-lucent, line 115: Labels merge looks good. One extra thought: if callers accidentally pass `hautech.ai/platform` in `opts.labels`, our merge will override it with the requested platform (desired). Consider a short comment here to document that behavior.

apps/server/src/entities/containerProvider.entity.ts
```diff
@@ -44,13 +50,47 @@ export class ContainerProviderEntity {
...
+    if (container && requestedPlatform) {
+      try {
+        const containerLabels = await this.containerService.getContainerLabels(container.id);
+        const existingPlatform = containerLabels?.[PLATFORM_LABEL];
+        if (!existingPlatform || existingPlatform !== requestedPlatform) {
```
noa-lucent, line 61: Non-reuse logic: This correctly handles mismatch and missing label when a platform is requested. Two edge cases to consider:
- When no platform requested: existing container should be reused even if it has a platform label. Current code does that; great.
- If stop/remove fails, we fall back to creating new (since we set `container = undefined`). That could leave an orphan. Would you want to log a warning with the container id on failure to aid cleanup?

apps/server/src/services/container.service.ts
```diff
@@ -55,25 +57,27 @@ export class ContainerService {
     }
```
noa-lucent, line 58: Behavioral check: ensure we only pass `{ platform }` to `docker.pull` when defined. The current branch does that, and tests assert both paths. Nice.

apps/server/__tests__/container.service.platform.test.ts
```diff
@@ -0,0 +1,92 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import Docker from 'dockerode';
+import { ContainerService } from '../src/services/container.service';
+import { LoggerService } from '../src/services/logger.service';
+import { PLATFORM_LABEL } from '../src/constants';
+
+vi.mock('dockerode', () => {
+  class MockContainer {
+    start = vi.fn(async () => {});
```
noa-lucent, line 9: Tests use `any` types in the mock (e.g., `modem: any`, `cb: any`). Could we tighten types a bit to avoid `any`, e.g., use minimal interfaces or `unknown` with narrowing? Not critical for unit tests, but our repo guideline is to avoid `any`.

apps/server/__tests__/containerProvider.entity.platform.test.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { ContainerProviderEntity, ContainerProviderStaticConfigSchema } from '../src/entities/containerProvider.entity';
+import { ContainerService } from '../src/services/container.service';
+import { ContainerEntity } from '../src/entities/container.entity';
+import { PLATFORM_LABEL } from '../src/constants';
+
+class MockContainer extends ContainerEntity {
```
noa-lucent, line 7: In tests, `svc` is typed as `Partial<ContainerService>` and then casted to `any` in multiple places. Consider defining a minimal interface for the subset under test to avoid `any` casts. Example:

interface IContainerServiceSubset { findContainerByLabels: ...; start: ...; getContainerLabels: ... }

Then use that for typing and remove the `as any` casts.

apps/server/src/services/container.service.ts
```diff
@@ -83,17 +87,19 @@ export class ContainerService {
...
       : optsWithDefaults.env
         ? Object.entries(optsWithDefaults.env).map(([k, v]) => `${k}=${v}`)
         : undefined;
 
-    const createOptions: ContainerCreateOptions = {
+    type CreateOptsWithPlatform = ContainerCreateOptions & { name?: string; platform?: string };
+    const createOptions: CreateOptsWithPlatform = {
       Image: optsWithDefaults.image,
       name: optsWithDefaults.name,
+      platform: optsWithDefaults.platform,
```
noa-lucent, line 105: ContainerCreateOptions typing: If we augment dockerode types to include `platform?: string`, we can drop the local intersection type `& { name?: string; platform?: string }` and simplify.

apps/server/src/entities/containerProvider.entity.ts
```diff
@@ -16,13 +17,18 @@ export const ContainerProviderStaticConfigSchema = z
       .optional()
       .describe('Shell script (executed with /bin/sh -lc) to run immediately after creating the container.')
       .meta({ 'ui:widget': 'textarea', 'ui:options': { rows: 6 } }),
+    platform: z
+      .enum(['linux/amd64', 'linux/arm64'])
+      .optional()
+      .describe('Docker platform selector for the workspace container')
+      .meta({ 'ui:widget': 'select' }),
```
noa-lucent, line 24: Schema/UI: Nice use of .describe. To ensure the UI schema shows a select, consider adding `.meta({ 'ui:widget': 'select' })` so consumers render a dropdown for the enum.

docs/technical-overview.md
```diff
@@ -213,6 +213,16 @@ Important config files
 - apps/server/src/services/config.service.ts: config schema and env mapping.
 - docker-compose.yml: local Mongo replica set bootstrap.
 
+Workspace container platform
```
noa-lucent, line 54: Docs: Great coverage. Could we add one line noting that cross-arch pulls require binfmt/qemu on Linux hosts, and that Desktop usually sets this up automatically?

apps/server/src/services/container.service.ts
noa-lucent, line 1: Enhancement suggestion re ensureImage: When a platform is specified, consider forcing a pull with `{ platform }` even if the tag is already present locally. The current early-return on `getImage(image).inspect()` may skip pulling the correct platform variant for multi-arch images. Safer behavior is:
- If platform provided, always invoke `docker.pull(image, { platform })` (or inspect manifests to verify the variant is present).

apps/server/src/entities/containerProvider.entity.ts
```diff
@@ -44,13 +50,47 @@ export class ContainerProviderEntity {
   async provide(threadId: string) {
     const labels = this.idLabels(threadId);
     let container: ContainerEntity | undefined = await this.containerService.findContainerByLabels(labels);
+
+    // Enforce non-reuse on platform mismatch if a platform is requested now
+    const requestedPlatform = this.cfg?.platform ?? this.opts.platform;
```
noa-lucent, line 56: Robustness: If stop() throws (e.g., already stopped or transient error), we skip removal due to the try/catch. Consider handling stop() errors but still attempting `remove(true)` to avoid orphans:

try { await container.stop(); } catch (e) { logger.warn(...) }
try { await container.remove(true); } catch (e) { logger.warn(...) }

Then proceed to recreate.

noa-lucent (commented):
Follow-up review after updates: great work addressing nearly all points. I found two remaining items before we can merge:

noa-lucent (commented):
Remaining deltas:
1) PLATFORM_LABEL: constants file seems missing from the PR; please add apps/server/src/constants.ts exporting PLATFORM_LABEL and fix imports.
2) Optional: If needed by UI, confirm toJSONSchema renders enum as select; the added meta hint is fine and should not break toJSONSchema.

Everything else in the checklist looks good: type-safety improvements applied, provider robustness implemented with 304/404 swallow, labels merge preserved and tested, CI-safe tests cover all paths, indentation and style fixed, and docs accurately describe behavior and caveats.

Once the constants file is added (or imports updated if path differs), I’ll approve.

noa-lucent (commented):
Addressed review feedback:

1) Type-safety
- Added TypeScript module augmentation in apps/server/src/types.d.ts to type dockerode options: `ImageCreateOptions.platform` and `ContainerCreateOptions.platform|name`.
- Removed casts/any usage in ContainerService. Platform is plumbed via typed options.

2) Stop/remove robustness and logging
- Provider now performs stop/remove in independent try/catch blocks, logs warn messages with short container id on failures, and still proceeds to remove if stop fails.
- Added explicit info logs when skipping reuse due to platform mismatch or when label is missing, including requested vs existing and cid.

3) Platform plumbing hygiene
- Only includes `platform` in create options when defined (avoid `?platform=undefined`).
- Introduced `PLATFORM_LABEL = 'hautech.ai/platform'` constant in ContainerService and used everywhere; comment notes we override a conflicting user label intentionally.

4) Safer ensureImage
- When `platform` is provided, we now always call `docker.pull(image, { platform })` even if `getImage(image).inspect()` succeeds, to ensure the correct variant is present. If `platform` is not provided and the image already exists, we skip pulling.

5) Tests and nits
- Tests updated to avoid broad `any` in mocks; use unknown + narrowings and minimal typed stubs; formatting tweaks.

All tests pass locally (Vitest). Please take another look; if this satisfies the feedback I can close duplicate PR #72 in favor of #71. Thanks!

noa-lucent:
Requesting review for platform support in Workspace (linux/amd64, linux/arm64) with non-reuse on mismatch. Maintainers, please review PR #71 at your earliest convenience.

apps/server/src/services/container.service.ts
noa-lucent: Nit: `query` is computed but never used. If the intent is to pass `platform` as a query param to Docker, wire it into the createContainer call (see next comment); otherwise remove this variable.

apps/server/src/services/container.service.ts
```diff
@@ -105,13 +111,18 @@ export class ContainerService {
       Tty: optsWithDefaults.tty ?? false,
       AttachStdout: true,
       AttachStderr: true,
-      Labels: optsWithDefaults.labels,
+      Labels: {
+        ...(optsWithDefaults.labels || {}),
+        ...(optsWithDefaults.platform ? { [PLATFORM_LABEL]: optsWithDefaults.platform } : {}),
+      },
```
noa-lucent, line 120: We currently cast `createOptions as unknown as ContainerCreateOptions` to satisfy types for the `platform` field. Dockerode supports passing query-string opts to createContainer; to avoid unsafe casts, consider one of:
- Narrow a local type that is assignable to ContainerCreateOptions and call `docker.createContainer(createOptions, { platform })` if your dockerode version exposes the second-arg opts; or
- Extend the Docker type locally with an overload for `createContainer(opts: ContainerCreateOptions, query?: { platform?: string })` and implement without the `unknown` cast; or
- Omit `platform` from createOptions and pass only as query param, depending on dockerode version.

Either way, let's remove the double cast and rely on proper types to keep us free of `any`/casts.

apps/server/src/services/container.service.ts
```diff
@@ -105,13 +111,18 @@ export class ContainerService {
       Tty: optsWithDefaults.tty ?? false,
       AttachStdout: true,
       AttachStderr: true,
```
noa-lucent, line 116: Bug: Platform is not actually sent as a query parameter to Docker. You compute `query` but never pass it, and `platform` is currently added to the create body. Docker Engine expects `platform` as a query string on POST /containers/create. Please update to call `docker.createContainer(createOptions, { platform: optsWithDefaults.platform })` (or the equivalent for your dockerode version) so the engine honors the platform. Update tests accordingly to assert the second arg contains `{ platform }`.

apps/server/src/services/container.service.ts
```diff
@@ -55,25 +57,27 @@ export class ContainerService {
...
             resolve();
           },
-          (event: any) => {
+          (event: { status?: string; id?: string }) => {
             if (event?.status && event?.id) {
               this.logger.debug(`${event.id}: ${event.status}`);
             } else if (event?.status) {
               this.logger.debug(event.status);
             }
           },
```
noa-lucent, line 78: For `ensureImage`, we avoid `any` by typing the followProgress event as `{ status?: string; id?: string }` which is good. However, there's still an unsafe cast in `({ platform } as { platform?: string })`. You can eliminate the cast by defining a small local type for the dockerode pull options you use, e.g. `type PullOpts = { platform?: string };` and then call `this.docker.pull(image, platform ? ({ platform } as PullOpts) : undefined, cb);` Or declare an overload signature via interface merging for Docker to include the typed opts. The goal is to avoid `unknown`/`any`/unnecessary casts in our codebase.

apps/server/src/entities/containerProvider.entity.ts
noa-lucent, line 1: Avoid `any` in error catches. Here and below we use `catch (e: any)`. Prefer `catch (e: unknown)` with a small type guard to read `statusCode` safely:

function statusCode(err: unknown): number | undefined {
  return typeof err === 'object' && err !== null && 'statusCode' in err
    ? (err as { statusCode?: number }).statusCode
    : undefined;
}

Then check `const code = statusCode(e); if (code !== 304 && code !== 404) throw e;` This aligns with our no-`any` rule and avoids broad casts.

apps/server/__tests__/container.service.platform.test.ts
noa-lucent, line 1: Tests: `container.service.platform.test.ts` spies only on the first arg to `createContainer`. Please extend the mock/spy to assert that the second argument (query) includes `{ platform: 'linux/arm64' }` when platform is set, and is absent when undefined. This will catch regressions where platform isn't threaded to the query string.

apps/server/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+// Centralized labels and shared constants for server
+export const PLATFORM_LABEL = 'hautech.ai/platform';
```
noa-lucent, line 7: Minor: Please export the platform enum as a reusable type to avoid string duplication. For example, in constants.ts:

export const SUPPORTED_PLATFORMS = ['linux/amd64','linux/arm64'] as const;
export type Platform = typeof SUPPORTED_PLATFORMS[number];
export const PLATFORM_LABEL = 'hautech.ai/platform';

Then reference `Platform` in ContainerOpts and zod: `z.enum(SUPPORTED_PLATFORMS as [Platform, ...Platform[]])`. This keeps the allowed values consistent across code and schema.

docs/README.md
```diff
@@ -7,5 +7,5 @@
 
 ## Recent Additions
```
noa-lucent, line 9: Docs: Great coverage. Please add a short note that when `platform` is undefined we do not set the `hautech.ai/platform` label at all (to preserve back-compat), and that reuse behavior remains unchanged. This matches the implementation and helps users reason about labels they may see.

docs/technical-overview.md
```diff
@@ -213,6 +213,16 @@ Important config files
...
+- containerProvider.staticConfig.platform: Optional; enum of `linux/amd64` or `linux/arm64`.
+- Behavior: When set, `docker pull` includes the platform selector and `docker.createContainer` receives `platform` as a query parameter. New containers are labeled with `hautech.ai/platform`.
```
noa-lucent, line 56: Docs: Consider explicitly stating that we will stop and remove an existing container on mismatch/missing platform label, with benign error handling for 304/404 during stop/remove. This mirrors the code and sets expectation for logs and behavior.

noa-lucent (commented):
Thanks for the well-scoped PR — functionality and tests are close. I verified each requested area and left line comments. A few correctness and type-safety issues remain before merge.

Requested changes
1) CreateContainer platform threading (bug)
- We compute `query` but never pass it, and currently put `platform` on the create body. Docker Engine expects `platform` as a query param on POST /containers/create. Please update to call `docker.createContainer(createContainerOpts, { platform })` (or the dockerode-appropriate equivalent) and remove the unused `query` variable. Update tests to assert the second arg contains `{ platform }`.

2) Avoid unsafe casts / any
- container.service.ts: Remove `as unknown as ContainerCreateOptions` and the cast on pull opts. Prefer typed overloads or a small local type for `{ platform?: string }`, or module augmentation for dockerode types so we can call without casts.
- containerProvider.entity.ts: Replace `catch (e: any)` with `catch (e: unknown)` and add a small type guard to read `statusCode` safely.
- tests: Replace `any` usages in mocks (e.g., `_opts: any`, `(...args: any[])`) with minimal typed interfaces or unknown + narrowing. Also avoid `(svc as any).docker`; use a public getter (e.g., `svc.getDocker()`) or expose a helper in the service for test access.

3) ensureImage platform correctness
- If `platform` is specified, do not early-return just because `getImage(image).inspect()` succeeds. Always call `docker.pull(image, { platform })` in that case to ensure the correct arch variant is available. Add a unit test simulating the image existing locally while `platform` is provided and assert that pull is still invoked with `{ platform }`.

4) Platform constant/type reuse
- In `apps/server/src/constants.ts`, export SUPPORTED_PLATFORMS and a `Platform` type. Use that type in `ContainerOpts` and in the zod schema (via `z.enum(SUPPORTED_PLATFORMS as [Platform, ...Platform[]])`) to avoid string duplication and keep schema/code in sync.

5) Tests improvements
- container.service.platform.test.ts: Extend spy to assert the second arg to `createContainer` includes `{ platform }` when set, and is absent when undefined. Also prefer a real PassThrough stream instead of `{} as unknown as ReadableStream` to avoid casts.
- containerProvider.entity.platform.test.ts: Consider an assertion that remove() is not called in the reuse-when-undefined case for completeness.

6) Docs
- README and technical-overview are solid. Please explicitly call out that when `platform` is undefined we do not set the `hautech.ai/platform` label (back-compat) and that stop/remove on mismatch swallows benign 304/404 errors.

Once these are addressed, I’m happy to re-review and approve. Nice work on threading, reuse policy, and CI-safe tests so far.

noa-lucent (commented):
Pushed follow-up changes addressing the review:

- Pass platform as a real top-level createContainer option so dockerode forwards it as a query param; removed unused query object.
- Types tightened: introduced `Platform` and `SUPPORTED_PLATFORMS`; added `PullOpts` and narrowed followProgress event typing; removed unsafe casts/double-casts; replaced `catch (any)` with `catch (unknown)` and a small `getStatusCode()` helper in provider.
- ensureImage now pulls even if image exists when a platform is specified to ensure the correct arch variant is present.
- Tests: use PassThrough for followProgress; added a test to assert pull happens even if inspect() succeeds when platform is set; removed anys from mocks and added an assertion that remove() is not called when platform is undefined and container is reused.
- Docs: clarified that when platform is undefined we do not set `hautech.ai/platform` and reuse behavior remains unchanged; noted that on mismatch/missing label we stop/remove and swallow benign 304/404 errors.

Let me know if you want me to split anything or adjust naming. Ready for re-review.

Issue #175, opened 2025-10-15
Phase 1: Node lifecycle adoption; pure templates; migrate core nodes (no back-compat)


rowan-stein:
[...]

PR #177, opened 2025-10-15
Feat/ben

Benkovichnikita:
Phase 1: Node lifecycle adoption; pure templates; migrate core nodes (no back-compat)

Implements Issue https://github.com/HautechAI/agents/issues/175.

Summary

Introduce NodeLifecycle interface (configure/start/stop/delete) with idempotency and allowed-states JSDoc.
LiveGraphRuntime refactor:
New helpers: startNode(id), stopNode(id), configureNode(id, cfg), deleteNode(id).
instantiateNode(): applies configure(config) if present; no auto-start.
disposeNode(): calls instance.delete() only; remove legacy destroy/close/stop fallbacks.
Remove legacy pause/resume/provision/deprovision helpers from runtime.
Add minimal getNodeStatus() bridge returning provisionStatus and dynamicConfigReady for UI.
Templates: make factories pure (construct-only); remove lifecycle side effects.
Core nodes migrated (MVP set):
SlackTrigger: setConfig -> configure; start/stop delegate to provision/deprovision; keep internal Provisionable semantics.
PRTrigger: implement start/stop around internal polling; provision hooks used internally; no dynamic config.
LocalMCPServer: setConfig -> configure; start/stop delegate to provision/deprovision; delete() final cleanup; discoverTools/callTool per-thread w/ threadId.
SimpleAgent: setConfig -> configure; runtime config updates applied; lifecycle start/stop/delete follow-ups remain (heavy init to move to start()).
Server routes:
Add /graph/nodes/:id/start, /stop, /configure, DELETE /graph/nodes/:id.
Add /graph/nodes/:id/dynamic-config/schema endpoint (read-only schema).
ReadinessWatcher emits node_status after readiness; uses runtime.getNodeStatus bridge.
UI:
Update hooks and panels to use start/stop actions instead of provision/deprovision and remove pause/resume actions.
API: add startNode/stopNode/configureNode/deleteNode helpers; dynamic-config schema loader supports new path.
Tests:
Update targeted tests to use configure/start/stop where applicable (SlackTrigger, LocalMCPServer, runtime helpers, dynamic config apply, MCP lifecycle changes, UI optimistic/actions tests).
Broad test suite alignment remains; many tests still reference legacy setConfig/pause/resume and external deps make local runs heavy.
Notes/Assumptions

No runtime back-compat: removed legacy pause/resume/provision/deprovision surfaces at runtime level; nodes may internally keep Provisionable semantics.
Templates are pure: no lifecycle calls inside factories.
getNodeStatus retained temporarily to keep UI working until a new status surface is designed; only returns provisionStatus and dynamicConfigReady.
Follow-ups/TODOs (Phase 1+)

Complete SimpleAgent lifecycle: move heavy init/graph compilation to start(); implement stop() teardown and delete() cleanup.
Replace remaining server/UI references to legacy actions where found (pause/resume remnants) and remove pausable capability display once not needed.
Broaden test migration: replace setConfig with configure; update helper types; align mocks for new lifecycle; trim tests needing external deps in CI or mock them.
Revisit ReadinessWatcher and status semantics with new lifecycle; consider richer status endpoint.
Run CI and fix any remaining path/import issues; ensure workspaces resolve dev deps for tests (zod, langchain, mongodb, etc.).
Breaking changes

Nodes must implement NodeLifecycle: configure/start/stop/delete. setConfig removed from runtime expectations.
Runtime no longer exposes pause/resume/provision/deprovision helpers; routes switched to start/stop/configure/delete.
Please review the runtime, server routes, and UI action changes. I will continue migrating remaining nodes/tests in subsequent PRs or commits under this issue.

### DIFF
--- a/apps/server/src/agents/simple.agent.ts
+++ b/apps/server/src/agents/simple.agent.ts
@@ -427,8 +427,8 @@ export class SimpleAgent extends BaseAgent {
    * Dynamically set configuration values like the system prompt.
    */
   // Overload preserves BaseAgent signature while exposing a more precise config shape for callers.
-  setConfig(config: Partial<SimpleAgentStaticConfig> & Record<string, unknown>): void;
-  setConfig(config: Record<string, unknown>): void {
+  configure(config: Partial<SimpleAgentStaticConfig> & Record<string, unknown>): void;
+  configure(config: Record<string, unknown>): void {
     const parsedConfig = SimpleAgentStaticConfigSchema.partial().parse(
       Object.fromEntries(
         Object.entries(config).filter(([k]) =>
@@ -515,7 +515,7 @@ export class SimpleAgent extends BaseAgent {
     // Attempt to call stop/destroy lifecycle if available
     const anyServer: any = server;
     try {
-      if (typeof anyServer.destroy === 'function') await anyServer.destroy();
+      if (typeof anyServer.delete === 'function') await anyServer.delete();
       else if (typeof anyServer.stop === 'function') await anyServer.stop();
     } catch (e) {
       const msg = e instanceof Error ? `${e.name}: ${e.message}` : String(e);
--- a/apps/server/src/graph/liveGraph.manager.ts
+++ b/apps/server/src/graph/liveGraph.manager.ts
@@ -13,6 +13,7 @@ import { Errors } from './errors';
 import { PortsRegistry } from './ports.registry';
 import { TemplateRegistry } from './templateRegistry';
 import type { NodeLifecycle } from '../nodes/node.types';
+import type { ProvisionStatus } from './capabilities';
 
 const configsEqual = (a: unknown, b: unknown) => JSON.stringify(a) === JSON.stringify(b); // unchanged
 
@@ -86,6 +87,22 @@ export class LiveGraphRuntime {
   }
   async deleteNode(id: string): Promise<void> {
     await this.disposeNode(id);
+  }
+
+  // Status helper retained for UI/diagnostics (no pause fallback)
+  getNodeStatus(id: string): { provisionStatus?: ProvisionStatus; dynamicConfigReady?: boolean } {
+    const inst = this.state.nodes.get(id)?.instance as unknown;
+    const out: { provisionStatus?: ProvisionStatus; dynamicConfigReady?: boolean } = {};
+    if (inst) {
+      const has = (name: string) => typeof (inst as any)[name] === 'function';
+      if (has('getProvisionStatus')) {
+        try { out.provisionStatus = (inst as any).getProvisionStatus(); } catch {}
+      }
+      if (has('isDynamicConfigReady')) {
+        try { out.dynamicConfigReady = !!(inst as any).isDynamicConfigReady(); } catch {}
+      }
+    }
+    return out;
   }
 
   private async _applyGraphInternal(next: GraphDefinition): Promise<GraphDiffResult> {
--- a/apps/server/src/graph/liveGraph.manager.ts
+++ b/apps/server/src/graph/liveGraph.manager.ts
@@ -12,7 +12,7 @@ import { LoggerService } from '../services/logger.service';
 import { Errors } from './errors';
 import { PortsRegistry } from './ports.registry';
 import { TemplateRegistry } from './templateRegistry';
-import type { Pausable, ProvisionStatus, Provisionable, DynamicConfigurable } from './capabilities';
+import type { NodeLifecycle } from '../nodes/node.types';
 
 const configsEqual = (a: unknown, b: unknown) => JSON.stringify(a) === JSON.stringify(b); // unchanged
 
@@ -26,8 +26,7 @@ export class LiveGraphRuntime {
     lastGraph: undefined,
   };
 
-  // Track paused state for nodes that don't implement isPaused()
-  private pausedFallback = new Set<string>();
+  // Simplified lifecycle: no paused/provision fallbacks
 
   private applying: Promise<unknown> = Promise.resolve(); // serialize updates
   private portsRegistry: PortsRegistry;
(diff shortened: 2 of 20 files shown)


apps/server/src/index.ts
noa-lucent, line 1: [MAJOR] GET /graph/nodes/:nodeId/status route removed
Issue: The UI/API still calls GET /graph/nodes/:id/status (see apps/ui/src/lib/graph/api.ts getNodeStatus), and several tests rely on it. Removing this route is likely the reason server-tests are failing.
Rule: API/route consistency and safe deprecations — keep existing read endpoints or migrate all callers in the same PR.
Suggestion:
- Restore the status route for now, delegating to runtime.getNodeStatus(nodeId) to keep the UI and tests working.
- If you intend to remove it later, add a deprecation plan and migrate UI to a streaming-only status source.
Minimal patch:
```ts
fastify.get('/graph/nodes/:nodeId/status', async (req) => {
  const { nodeId } = req.params as { nodeId: string };
  return runtime.getNodeStatus(nodeId);
});
```
Rationale: Keeps read-side compatibility while you transition actions to start/stop/configure/delete.

apps/ui/src/lib/graph/api.ts
```diff
@@ -77,8 +77,11 @@ export const api = {
     // If plain object, validate it's likely a schema; otherwise null
     return isLikelyJsonSchemaRoot(data) ? (data as Record<string, unknown>) : null;
   },
-  postNodeAction: (nodeId: string, action: 'pause' | 'resume' | 'provision' | 'deprovision') =>
-    http<void>(`${BASE}/graph/nodes/${encodeURIComponent(nodeId)}/actions`, { method: 'POST', body: JSON.stringify({ action }) }),
+  startNode: (nodeId: string) => http<void>(`${BASE}/graph/nodes/${encodeURIComponent(nodeId)}/start`, { method: 'POST' }),
+  stopNode: (nodeId: string) => http<void>(`${BASE}/graph/nodes/${encodeURIComponent(nodeId)}/stop`, { method: 'POST' }),
+  configureNode: (nodeId: string, config: Record<string, unknown>) =>
+    http<void>(`${BASE}/graph/nodes/${encodeURIComponent(nodeId)}/configure`, { method: 'POST', body: JSON.stringify({ config }) }),
+  deleteNode: (nodeId: string) => http<void>(`${BASE}/graph/nodes/${encodeURIComponent(nodeId)}`, { method: 'DELETE' }),
```
noa-lucent, line 84: [MAJOR] UI still depends on GET /graph/nodes/:id/status
Issue: getNodeStatus still fetches /graph/nodes/:id/status, but the server removed this route in this PR. This will break runtime status view and tests.
Suggestion: Either (a) keep the status route on the server (preferred short-term), or (b) update UI to derive status exclusively from the websocket events and remove this fetch.
Given the current readinessWatcher usage, option (a) is the quickest to unblock CI.

apps/server/src/graph/liveGraph.manager.ts
```diff
@@ -328,26 +306,10 @@ export class LiveGraphRuntime {
...
-          if (typeof fn === 'function') {
-            try {
-              await (fn as Function).call(inst);
-            } catch {}
-            break;
-          }
-        }
-      }
+    // Call lifecycle teardown if present (delete)
+    const inst = live.instance as NodeLifecycle | undefined;
```
noa-lucent, line 310: [MAJOR] deleteNode/disposeNode calls only instance.delete() and removed fallbacks
Issue: The runtime used to call destroy/dispose/close/stop fallbacks on node removal. Now it only calls delete() if present. Nodes that haven’t been migrated will leak resources or fail to stop (e.g., instances that only implement stop() or legacy destroy()).
Suggestion: For Phase 1, keep a temporary fallback path until all nodes are migrated:
```ts
const instAny = live.instance as any;
if (typeof instAny.delete === 'function') await instAny.delete();
else if (typeof instAny.destroy === 'function') await instAny.destroy();
else if (typeof instAny.dispose === 'function') await instAny.dispose();
else if (typeof instAny.close === 'function') await instAny.close();
else if (typeof instAny.stop === 'function') await instAny.stop();
```
Also log a warning when falling back, so we can track remaining migrations.
Rationale: Avoids leaks and spurious test failures while migration is in progress.

apps/server/src/graph/liveGraph.manager.ts
```diff
@@ -170,7 +148,7 @@ export class LiveGraphRuntime {
       const live = this.state.nodes.get(nodeId);
       if (!live) continue;
       try {
```
noa-lucent, line 150: [MINOR] configureNode updates config store only after successful configure()
Consider ordering and error propagation:
- If configure() throws, the runtime shouldn’t persist the new config. Right now that is true (updateNodeConfig is after await). Good.
- Suggest adding a try/catch to surface a clear error with node id and partial config keys to server routes for better 500 messages.
Also, consider no-op when new cfg is deep-equal to existing (you already have configsEqual helper; you could short-circuit here to avoid unnecessary configure calls).

apps/server/src/triggers/slack.trigger.ts
noa-lucent, line 1: [MAJOR] Missing delete() lifecycle for full teardown
Issue: LiveGraphRuntime.disposeNode now only calls instance.delete(). SlackTrigger implements start/stop but not delete(), so removing a SlackTrigger node will skip teardown and likely leak the socket-mode client and listeners.
Suggestion: Implement delete() delegating to stop(), and clear any residual listeners/timers if applicable.
```ts
async delete(): Promise<void> { await this.stop(); }
```
Rationale: Ensures consistent cleanup on node removal under the new lifecycle and prevents resource leaks.

apps/server/src/triggers/pr.trigger.ts
noa-lucent, line 1: [MAJOR] Missing delete() lifecycle for full cleanup
Issue: Runtime now calls delete() when a node is removed. PRTrigger implements start/stop but no delete(). If polling timers or internal state need cleanup, it should implement delete() delegating to stop().
Suggestion:
```ts
async delete(): Promise<void> { await this.stop(); }
```
Rationale: Aligns with NodeLifecycle and prevents leakage of setTimeout handles.

apps/server/src/mcp/localMcpServer.ts
```diff
@@ -417,17 +415,14 @@ export class LocalMCPServer implements McpServer, Provisionable, DynamicConfigur
     }
   }
```
noa-lucent, line 417: [MINOR] Rename destroy() to delete(): ensure no external callers still use destroy()
Good: You’ve implemented delete() and updated SimpleAgent to call delete() first.
Risk: Any remaining external code/tests calling destroy() will now be no-ops. Consider keeping a temporary alias for back-compat:
```ts
async destroy(): Promise<void> { await this.delete(); }
```
This can be removed once all references are migrated.

apps/server/src/agents/simple.agent.ts
```diff
@@ -515,7 +515,7 @@ export class SimpleAgent extends BaseAgent {
     // Attempt to call stop/destroy lifecycle if available
     const anyServer: any = server;
     try {
```
noa-lucent, line 517: [MINOR] Fallback order: delete() then stop() — consider keeping destroy() alias
You updated to prefer delete() over stop(), which matches the new lifecycle. To smooth migration, consider supporting both delete() and legacy destroy() for a bit:
```ts
if (typeof anyServer.delete === 'function') await anyServer.delete();
else if (typeof anyServer.destroy === 'function') await anyServer.destroy();
else if (typeof anyServer.stop === 'function') await anyServer.stop();
```
Rationale: A few server-side components may still implement destroy() only.

apps/server/src/nodes/node.types.ts
noa-lucent, line 1: [MINOR] JSDoc: clarify idempotency semantics and expected error surface
- Consider documenting what errors should propagate vs. be swallowed by the runtime (e.g., configure validation errors should bubble up; stop() failing should log and continue?).
- Add a brief note on thread-safety expectations if nodes are interacted with concurrently via HTTP routes.
This will help implementers and test authors.

apps/server/src/templates.ts
```diff
@@ -177,11 +177,7 @@ export function buildTemplateRegistry(deps: TemplateRegistryDeps): TemplateRegis
       )
       .register(
         'slackTrigger',
-        () => {
-          const instance = new SlackTrigger(logger, vault);
-          void instance.start();
-          return instance;
-        },
+        () => new SlackTrigger(logger, vault),
         {
```
noa-lucent, line 181: [MINOR] Purified templates: good. Consider explicit comment noting "no auto-start" policy
Here you changed slackTrigger factory to return a new SlackTrigger without calling start(). Add a short comment near the registration explaining that factories must be pure and not invoke lifecycle methods.
This helps future contributors follow the convention.

apps/ui/src/components/graph/NodeDetailsPanel.tsx
noa-lucent, line 1: [MINOR] UI labels still mention pause/resume in comments
Consider removing stale comments referencing pause/resume to avoid confusion.
Also, NodeStatus still includes isPaused in types; if pause is deprecated, plan removal in a follow-up to reduce noise.

apps/server/src/index.ts
noa-lucent, line 1: [MAJOR] Missing GET /graph/nodes/:id/status route used by UI
Issue: UI/api.ts still calls GET /graph/nodes/:id/status, but this server no longer exposes it. This causes 404s and breaks useNodeStatus/react-query flows.
Rule: Keep temporary status surface consistent across server/UI during migration.
Suggestion:
Add a minimal bridge endpoint that proxies runtime.getNodeStatus(id):

```ts
fastify.get('/graph/nodes/:nodeId/status', async (req) => {
  const { nodeId } = req.params as { nodeId: string };
  return runtime.getNodeStatus(nodeId);
});
```
Rationale: Unblocks UI and tests until a richer status surface replaces it.

apps/server/src/graph/liveGraph.manager.ts
```diff
@@ -328,26 +306,10 @@ export class LiveGraphRuntime {
...
-              await (fn as Function).call(inst);
-            } catch {}
-            break;
-          }
-        }
-      }
+    // Call lifecycle teardown if present (delete)
+    const inst = live.instance as NodeLifecycle | undefined;
+    if (inst && typeof inst.delete === 'function') {
+      try { await inst.delete(); } catch {}
```
noa-lucent, line 312: [MAJOR] disposeNode(): delete-only teardown risks leaks for nodes lacking delete()
Issue: Only instance.delete() is called. Nodes like SlackTrigger/PRTrigger currently don’t implement delete(), so deleting the node won’t stop timers/sockets.
Rule: code-guide/typescript/0026-single-responsibility and 0027-side-effects-at-edges-core-logic-pure — lifecycles should be predictable and cleaned up.
Suggestion: Keep a safe fallback to stop():
```ts
if (inst && typeof inst.delete === 'function') {
  await inst.delete();
} else if (inst && typeof (inst as any).stop === 'function') {
  await (inst as any).stop();
}
```
Alternatively, implement delete() for all migrated nodes now.
Rationale: Prevents resource leaks and flakiness during graph updates.

noa-lucent (changes requested):
Overall direction looks good: unified NodeLifecycle, pure templates, and updated routes/hooks are clear improvements. A few issues need to be addressed before merge to keep UI and runtime behavior consistent and to avoid lifecycle-related leaks.

Blocking (MAJOR)
- Restore minimal status route: UI/api.ts still calls GET /graph/nodes/:id/status. Please add a small bridge on the server that returns runtime.getNodeStatus(id) to avoid 404s and breaking useNodeStatus.
- Runtime dispose cleanup: disposeNode currently calls only instance.delete(). Nodes like SlackTrigger and PRTrigger don’t implement delete(), so deleteNode/dispose will skip cleanup. Either:
  - Add a stop() fallback in disposeNode, or
  - Implement delete() = await stop() in all nodes now (SlackTrigger/PRTrigger at minimum).

Minor
- Document lifecycle purity in templates where you removed auto-start (comment/README note helps future contributors).
- UI: remove stale pause/resume comments and ensure NodeStatus typing reflects the new surface. Guard NodeDetailsPanel against missing status until the first socket event or the restored status route responds.
- hooks.ts: debounce map for dynamic-config invalidation — consider extracting to a tiny helper with a comment about socket + polling rationale.
- simple.agent.ts: migration shim for delete/destroy/stop fallback could be helpful during the transition.

Tests
- Add/adjust tests to cover the restored status endpoint or update UI mocks accordingly.

Once the status bridge and lifecycle tear-down are aligned, the rest looks solid for Phase 1. See inline comments for details and suggested snippets.

PR #322, opened 2025-10-18
Fix: SimpleAgent lifecycle config propagation (epic #248)

### LINKED ISSUE #317: Phase 3 redux: Add regression test for config loss (commit 041a783) and fix SimpleAgent lifecycle propagation
Context
- Commit 041a7836a083cfcd2e5d27b3ea2aa9df1dce15be (Phase 3) introduced a regression where SimpleAgent loses configured values (e.g., systemPrompt, model) when created via lifecycle (configure/start).

Task
- Add a regression test demonstrating that a simpleAgent node with provided static config (e.g., systemPrompt, model, summarization fields, restriction flags) retains those values after LiveGraphRuntime.apply().
- Fix SimpleAgent lifecycle so configure/start correctly apply provided static config without injecting defaults or dropping keys. No changes to lgnodes.

Acceptance
- New tests under apps/server/__tests__ fail on 041a783 and pass with the fix:
  - When graph config specifies systemPrompt and model, the agent instance reflects these values (callModelNode.systemPrompt, llm.model).
  - Updating node config via runtime.apply updates these values live.
- CI green. Single PR with tests+fix. No lgnodes changes.

Links
- Epic: #248
- Regressing commit: 041a7836a083cfcd2e5d27b3ea2aa9df1dce15be}

rowan-stein:
Adds a regression test and fixes SimpleAgent lifecycle propagation:

- Regression test: via LiveGraphRuntime + templates, verifies provided systemPrompt/model (and related fields) are applied on start and updated on configure.
- Fix: SimpleAgent.configure/start now properly propagate provided config without injecting defaults or dropping keys; idempotent lifecycle.

Closes #317. Links epic #248.

### DIFF
--- a/apps/server/__tests__/live.graph.runtime.simpleAgent.config.propagation.test.ts
+++ b/apps/server/__tests__/live.graph.runtime.simpleAgent.config.propagation.test.ts
@@ -1,0 +1,110 @@
+import { describe, it, expect, vi } from 'vitest';
+import { LiveGraphRuntime } from '../src/graph/liveGraph.manager';
+import { buildTemplateRegistry } from '../src/templates';
+import { LoggerService } from '../src/services/logger.service';
+import { ContainerService } from '../src/services/container.service';
+import { ConfigService } from '../src/services/config.service';
+import { CheckpointerService } from '../src/services/checkpointer.service';
+
+// Avoid any real network calls by ensuring ChatOpenAI token counting/invoke are not used in this test.
+// We don't invoke the graph; we only verify propagation of config to internal nodes/fields.
+
+describe('LiveGraphRuntime -> SimpleAgent config propagation', () => {
+  function makeRuntime() {
+    const logger = new LoggerService();
+    const containerService = new ContainerService(logger);
+    const configService = new ConfigService({
+      githubAppId: 'test',
+      githubAppPrivateKey: 'test',
+      githubInstallationId: 'test',
+      openaiApiKey: 'test',
+      githubToken: 'test',
+      mongodbUrl: 'mongodb://localhost:27017/?replicaSet=rs0',
+    } as any);
+    const checkpointerService = new CheckpointerService(logger);
+    // Patch to bypass Mongo requirement for this lightweight test
+    (checkpointerService as any).getCheckpointer = () => ({
+      async getTuple() { return undefined; },
+      async *list() {},
+      async put(_config: any, _checkpoint: any, _metadata: any) { return { configurable: { thread_id: 't' } } as any; },
+      async putWrites() {},
+      getNextVersion() { return '1'; },
+    });
+    const mongoService = { getDb: () => ({} as any) } as any;
+    const registry = buildTemplateRegistry({ logger, containerService, configService, checkpointerService, mongoService });
+    const runtime = new LiveGraphRuntime(logger, registry);
+    return { runtime };
+  }
+
+  it('applies provided config on configure/start and updates on re-apply', async () => {
+    const { runtime } = makeRuntime();
+    const systemPrompt = 'You are Strict.';
+    const model = 'gpt-9-test';
+    const keep = 123;
+    const max = 456;
+    const restrict = true;
+    const restrictionMessage = 'Always call a tool first.';
+
+    const graph1 = {
+      nodes: [
+        {
+          id: 'agent',
+          data: {
+            template: 'simpleAgent',
+            config: {
+              systemPrompt,
+              model,
+              summarizationKeepTokens: keep,
+              summarizationMaxTokens: max,
+              restrictOutput: restrict,
+              restrictionMessage,
+            },
+          },
+        },
+      ],
+      edges: [] as any[],
+    };
+
+    await runtime.apply(graph1 as any);
+    const agent: any = runtime.getNodeInstance('agent');
+    expect(agent).toBeDefined();
+
+    // Validate propagation into internal nodes/fields
+    expect((agent as any).callModelNode?.['systemPrompt']).toBe(systemPrompt);
+    expect((agent as any).llm?.['model']).toBe(model);
+    expect((agent as any).summarizeNode?.['keepTokens']).toBe(keep);
+    expect((agent as any).summarizeNode?.['maxTokens']).toBe(max);
+    expect((agent as any).restrictOutput).toBe(restrict);
+    expect((agent as any).restrictionMessage).toBe(restrictionMessage);
+
+    // Update config live and re-apply
+    const newSystemPrompt = 'You are Even Stricter.';
+    const newModel = 'gpt-9x-test';
+    const graph2 = {
+      nodes: [
+        {
+          id: 'agent',
+          data: {
+            template: 'simpleAgent',
+            config: {
+              systemPrompt: newSystemPrompt,
+              model: newModel,
+              summarizationKeepTokens: keep,
+              summarizationMaxTokens: max,
+              restrictOutput: restrict,
+              restrictionMessage,
+            },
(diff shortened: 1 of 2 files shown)


noa-lucent (changes requested):
Summary
- Verified CI is green (lint, test, build for server/UI) and no lgnodes changes were introduced. The regression test captures Phase 3 config loss and the lifecycle refactor makes configure/start idempotent and applies only provided keys.

Blocking items
1) Avoid any and redundant casts in SimpleAgent
   - apps/server/src/agents/simple.agent.ts: In two places we do SimpleAgent.allowedConfigKeys.includes(k as any). Please replace the any with a precise key type. Example:
     type AllowedKey = keyof SimpleAgentStaticConfig;
     private static readonly allowedConfigKeys: readonly AllowedKey[] = [
       "title","model","systemPrompt","debounceMs","whenBusy","processBuffer","summarizationKeepTokens","summarizationMaxTokens","restrictOutput","restrictionMessage","restrictionMaxInjections",
     ];
     // Usage:
     const filtered = Object.fromEntries(
       Object.entries(config).filter(([k]) => (new Set<AllowedKey>(SimpleAgent.allowedConfigKeys)).has(k as AllowedKey)),
     );
     Similarly in configure(): filter against the Set<AllowedKey> without any.

   - apps/server/src/agents/simple.agent.ts: setConfig now casts to string (as string) when updating systemPrompt/model. Please avoid redundant casts by refining the guard instead:
     if ("systemPrompt" in filtered && typeof parsedConfig.systemPrompt === "string") {
       this.callModelNode.setSystemPrompt(parsedConfig.systemPrompt);
     }
     if ("model" in filtered && typeof parsedConfig.model === "string") {
       this.llm.model = parsedConfig.model;
     }

2) Tests: remove any and unused import
   - apps/server/__tests__/live.graph.runtime.simpleAgent.config.propagation.test.ts:
     • vi is imported but unused; please remove it.
     • Replace as any usages with typed alternatives. Suggestions:
       - ConfigService init: use a typed object instead of as any, e.g.:
         const cfg = {
           githubAppId: "test",
           githubAppPrivateKey: "test",
           githubInstallationId: "test",
           openaiApiKey: "test",
           githubToken: "test",
           mongodbUrl: "mongodb://localhost:27017/?replicaSet=rs0",
         } satisfies ConstructorParameters<typeof ConfigService>[0];
         const configService = new ConfigService(cfg);
       - Checkpointer patch: introduce a minimal type and spy/mocks instead of as any. For example with Vitest:
         const fake = { getTuple: async () => undefined, async *list() {}, put: async () => ({ configurable: { thread_id: "t" } }), putWrites: async () => {}, getNextVersion: () => "1" };
         vi.spyOn(checkpointerService, "getCheckpointer").mockReturnValue(fake);
         If a cast is unavoidable, prefer a single unknown as ConcreteType rather than any, and avoid double assertions.
       - Graph shapes: avoid edges: [] as any[] and runtime.apply(graph as any). Prefer a typed/satisfies shape for the graph object if a public type is available; alternatively use satisfies for the node/edge collections.
       - Access to private fields via any for assertions is brittle. Consider exposing read-only getters or a debug snapshot on SimpleAgent for test introspection (e.g., getModelId(), getSystemPrompt(), getSummarizationOptions()) so tests don’t need to subvert visibility.

Nice-to-have (non-blocking)
- setConfig() uses .passthrough() on the schema while already filtering to allowed keys; dropping passthrough would simplify the mental model and reduce risk of accepting unknowns.
- In configure(), delta comparison can use Object.is(before, v) for primitives rather than JSON.stringify for clarity and perf.

Functionality verification
- Provided keys are applied on start and updated on configure; start is idempotent and the node instance is preserved across apply() calls. No defaults appear to be injected through start/configure. CI passes.

Please address the blocking items; once those are fixed, this is good to merge.}

noa-lucent (changes requested):
Re-review summary
- Thanks for the quick follow-up. The lifecycle typing changes look much better: typed AllowedKey set, snapshot is Partial<SimpleAgentStaticConfig>, and passthrough() removed. Logic still satisfies: no defaults injected, provided keys applied on start, updates applied on configure, start is idempotent, and no lgnodes changes. CI is currently running (Lint in progress at time of review).

Remaining blocking items
1) Tests: any usage and double assertions
   File: apps/server/__tests__/live.graph.runtime.simpleAgent.config.propagation.test.ts
   - new ConfigService({...} as any): please eliminate any. Suggested:
     const cfg = {
       githubAppId: "test",
       githubAppPrivateKey: "test",
       githubInstallationId: "test",
       openaiApiKey: "test",
       githubToken: "test",
       mongodbUrl: "mongodb://localhost:27017/?replicaSet=rs0",
     } satisfies ConstructorParameters<typeof ConfigService>[0];
     const configService = new ConfigService(cfg);
   - (checkpointerService as any).getCheckpointer = () => ({ ... }): replace the any cast and the internal any parameters. Prefer a typed fake and spy:
     const fake = { getTuple: async () => undefined, async *list() {}, put: async () => ({ configurable: { thread_id: "t" } }), putWrites: async () => {}, getNextVersion: () => "1" };
     vi.spyOn(checkpointerService, "getCheckpointer").mockReturnValue(fake as ReturnType<typeof checkpointerService.getCheckpointer>);
     Ensure method params are correctly typed if you keep an inline object (avoid _config: any, etc.).
   - mongoService: avoid double assertion (as unknown as MongoService). This violates our “Never use double unsafe type assertions” policy. Options:
     • Define a minimal class that implements MongoService with getDb():
       class TestMongoService implements MongoService { getDb() { return {} as any; } }
       const mongoService = new TestMongoService();
       // If a cast is unavoidable, use a single cast (as MongoService), not a double.
   - Agent access: const agent: any = runtime.getNodeInstance(agent); Please avoid any. If there is no public debug surface, you can introduce a tiny test-only interface and cast once:
     type InspectableAgent = { callModelNode?: { systemPrompt?: string }; llm?: { model?: string }; summarizeNode?: { keepTokens?: number; maxTokens?: number }; restrictOutput?: boolean; restrictionMessage?: string };
     const agent = runtime.getNodeInstance(agent) as unknown as InspectableAgent;
     This keeps a single safe-ish assertion and avoids pervasive any.

2) SimpleAgent: remove redundant casts to string
   File: apps/server/src/agents/simple.agent.ts
   - setConfig(): for systemPrompt/model, the schema narrows to string|undefined. Guard and assign without casting:
     if (Object.prototype.hasOwnProperty.call(filtered, "systemPrompt") && parsedConfig.systemPrompt !== undefined) {
       this.callModelNode.setSystemPrompt(parsedConfig.systemPrompt);
     }
     if (Object.prototype.hasOwnProperty.call(filtered, "model") && parsedConfig.model !== undefined) {
       this.llm.model = parsedConfig.model;
     }

Non-blocking suggestions
- You can extract a type guard for keys to avoid the cast in Set.has:
  type AllowedKey = typeof SimpleAgent.allowedConfigKeysArray[number];
  const isAllowedKey = (k: string): k is AllowedKey => SimpleAgent.allowedConfigKeys.has(k as AllowedKey);
  // then: Object.entries(config).filter(([k]) => isAllowedKey(k))
- In configure(), consider Object.is for primitive comparisons over JSON.stringify.

CI
- CI rerun is in progress following the latest commit. I will re-check once it completes.}

noa-lucent (dismissed):
Looks great. Final pass:
- Tests no longer use any/double assertions and have typed fakes/spies
- SimpleAgent guards refined; redundant casts removed
- Lifecycle behavior remains idempotent and correctly propagates provided config without defaults; no lgnodes changes

CI was running at the time of approval; please ensure green before merge.

PR #306, opened 2025-10-18
feat(server): Install Nix packages on container startup

rowan-stein:
Implements issue #304.\n\n- Adds installNixPackages() to ContainerProviderEntity\n- On container creation and reuse, attempts to run: `nix profile install 'github:NixOS/nixpkgs/<COMMIT>#<ATTR>' ...` for each configured package in cfg.nix.packages.\n- Skips gracefully when nix is not present in the container (logs error).\n- Reuses ncps via existing NIX_CONFIG injection.\n\nIncludes best-effort logging and 15 min timeout for the installation step.

Issue #304, opened 2025-10-18
Install Nix packages on container startup


vitramir:
This is 3rd phase of nix package manager integration. There is already UI to add nix packages to workspace and packages are stored in node state with persistency in graph. There is caching layer with ncps. Now need to implement last step and add command to install packages in the container. 

use this command to install packages:
```
nix profile install \
  'github:NixOS/nixpkgs/c12c63cd6c5eb34c7b4c3076c6a99e00fcab86ec#nodejs_24' \
  'github:NixOS/nixpkgs/c12c63cd6c5eb34c7b4c3076c6a99e00fcab86ec#git' \
  'github:NixOS/nixpkgs/<COMMIT_HASH>#<ATTRIBUTE_PATH>'
```

COMMIT_HASH and ATTRIBUTE_PATH are stored in packages array in node

Operation may fail in the container if there is no nix. Handle this error and log error message.

PR #332, opened 2025-10-19
feat(server): Install Nix packages on container startup

rowan-stein:
Supersedes #306. Implements #304.

Summary
- Install Nix packages on container startup (fresh and reuse) using canonical resolved nix shape
- Typed internal projection for cfg.nix; removed unsafe/double assertions and legacy aliases
- Idempotency via sentinel digest (/var/lib/hautech/<cid>/nix-specs.sha256) and per-container lock dir (/var/lock/hautech-nix-<cid>)
- Shell safety: prefer argv exec and strict input validation
- Observability: include container id, spec count, durations; improved error logs
- Timeout configurable via cfg.nix.timeoutSeconds (default 900s); documented in schema comments
- Tests: cover fresh/reuse, no-nix, non-zero exit, idempotency, concurrency, timeout

Notes
- External schema remains opaque to avoid UI breaks; server derives internal typed shape
- Behavior aligns with prior review guidance from #306

Please review with focus on idempotency, concurrency, safety, observability, and typing. CI is expected to pass.


noa-lucent (changes requested):
High-level review checklist
- Verified typed internal nix config projection and removal of unsafe assertions
- Checked idempotency (sentinel digest) and per-container locking
- Ensured argv exec usage and input validation
- Reviewed observability: container id, spec counts, durations, error logs
- Confirmed configurable timeout is applied consistently
- Reviewed tests for coverage and style conformance

Blocking issues and recommendations
1) Lock release on exceptions
   In ensureNixPackages, the precheck acquires a lock directory (exitCode 0), but the release (rmdir) happens only on normal paths. If any error throws after lock acquisition, the catch block logs and returns without releasing the lock, leaving a stale lock that will block all future installs.
   Action: Track a boolean lockAcquired after precheck succeeds and move the rmdir call into a finally block so it always runs when lockAcquired is true. Also keep the early-release path for the 42 (unchanged) branch.

2) Sentinel digest stability vs. order
   The digest is computed from refs.join("\n"). If the input order changes but the set is identical, digest changes and triggers unnecessary reinstalls.
   Action: Consider sorting refs (e.g., by `${commitHash}#${attributePath}`) before hashing, or explicitly document that order is significant and add a test for your chosen behavior. I recommend sorting for idempotency.

3) Input validation and shell safety
   You improved safety by using argv form with ["sh","-lc", ...], and by deriving a typed internal shape. To further harden:
   - Enforce commitHash to be 40 hex chars and attributePath to match a safe pattern (e.g., ^[-a-zA-Z0-9_.\/]+$) in ResolvedNixPackageItemSchema. If already present, great; otherwise please add.
   - Prefer direct argv invocation of nix: exec(["nix","profile","install", ...], { env: {...PATH...} }) instead of passing a shell string through sh -lc. You can set PATH via exec options (env) and remove PATH_PREFIX and most quoting concerns. This avoids potential injection via attributePath and reduces shell quoting pitfalls.

4) Minor duplication
   The code to derive rawArr/originalCount from cfg.nix.packages appears twice. Consider extracting a small helper that returns { resolved: InternalNixConfig["packages"] ?? [], originalCount } to keep the logic single-sourced with readNixInternal().

5) Tests: avoid any/double assertions and add coverage
   - Current tests use `as any` and `as unknown as ...`. Our enforced rules forbid using any and double assertions. Please refactor tests to avoid them:
     • For logger assertions, prefer using vi.spyOn on a real LoggerService instance or implement a StubLogger that collects messages in typed arrays and assert on those, avoiding `as any` casts.
     • For config, setConfig accepts Record<string, unknown>, so you can build plain objects without casting to ContainerProviderStaticConfig, or use `satisfies` where applicable.
   - Suggested additional tests:
     • Sentinel order invariance (if you adopt sorting) or “order significant” behavior (if you keep order) documented and tested.
     • Ensure that on a failure path after acquiring the lock, the lock is still released (once you move release to finally).
     • Ensure that sentinel is not written when per-package fallback has any failure (you already track allOk; add an assertion to tests).

6) Observability
   Great improvements: container id, spec count, duration_ms, explicit skip/in-progress logs. Consider adding the digest prefix (e.g., first 8 chars) to logs for easier cross-correlation if helpful.

Overall, the refactor meaningfully improves typing, idempotency, concurrency control, and telemetry. Addressing the lock-release bug and test type assertions are required before merge. I will re-check once updates are pushed and CI is green.}

PR #375, opened 2025-10-20
Migrate to OpenAI Responses API (non-streaming); prep tolerant parsing and stabilize tool binding

### LINKED ISSUE #374: Migrate to OpenAI Responses API (non-streaming) and add tolerant parser for new content types
Context
- Branch: feat/migrate-to-responses
- Goal: finalize migration from Chat Completions to OpenAI Responses API (non-streaming only) and ensure our app correctly handles new response item types.
- Current failure observed during testing:
  Item 'rs_…' of type 'reasoning' was provided without its required following item.
  This occurs when using useResponsesApi: true with @langchain/openai; the adapter enforces ordering for reasoning items.

Problem summary
- The app can send requests using the Responses API (via ChatOpenAI(useResponsesApi: true)), but parsing the returned content fails when a reasoning-capable model returns a reasoning item not immediately followed by an output_text item, or when tool-first responses appear. The error is thrown upstream, bubbling into our run.

Scope (non-streaming only)
- Complete the migration to Responses API for the main agent turn (CallModelNode), and add tolerant parsing for content item types: output_text, reasoning, tool_use, tool_result. Leave streaming out of scope.
- Keep SummarizationNode and MemoryCallModelNode on ChatOpenAI for now (no change required there). Token counting remains as-is.

Proposed approach
1) Introduce a direct Responses client and parser
   - New file: packages/platform-server/src/services/openai.responses.ts
     - Implement a small wrapper around the OpenAI SDK (openai@6.x) to call POST /v1/responses (non-streaming).
     - Map our internal messages to Responses input messages:
       • SystemMessage → role:"system", content:[{ type:"input_text", text }]
       • HumanMessage → role:"user", content:[{ type:"input_text", text }]
       • AIMessage (no tools) → role:"assistant", content:[{ type:"output_text", text }]
       • AIMessage with tool calls → role:"assistant", content:[ optional { type:"output_text", text }, then { type:"tool_use", id, name, input } for each call ]
       • ToolMessage → role:"tool", content:[{ type:"tool_result", tool_use_id:id, content:string | object }]
     - Tools schema: { type:"tool", name, description?, input_schema: JSON Schema object } (convert existing tool schema to JSON Schema; we already have Zod schemas on tools via LangChainToolAdapter.inferArgsSchema).
     - Parse response.output[] and within assistant message.content[]:
       • Collect output_text segments (concatenate with \n)
       • Collect tool_use calls → map to AIMessage.tool_calls with { id, name, args }
       • Collect reasoning segments into metadata (do not expose to user-facing content)
       • Be tolerant of ordering issues: if a reasoning item is not immediately followed by output_text, log a warning and continue (do not throw). Preserve subsequent tool_use items.
       • Skip unknown top-level output items or content item types with a debug log.
     - Return both text and toolCalls plus original response id/usage (if present) for tracing.

2) Wire CallModelNode to use the direct Responses client
   - File: packages/platform-server/src/lgnodes/callModel.lgnode.ts
     - Add a feature flag AGENTS_USE_DIRECT_RESPONSES (default true) coming from ConfigService.
     - When enabled: build Responses request payload from the last 10 messages (existing window) + optional memory injection; include tools; call the client; parse; and produce an LLMResponse with { raw, content, toolCalls } similar to today.
     - When disabled: fall back to current ChatOpenAI.bindTools(...).invoke(...) path.

3) Config plumbing
   - File: packages/platform-server/src/services/config.service.ts
     - Add optional env flag: AGENTS_USE_DIRECT_RESPONSES (string "1"/"true" → true). Default to true when not provided.
     - Ensure OPENAI_API_KEY and OPENAI_BASE_URL are plumbed to the Responses client.

4) Logging and error handling
   - Do not throw on content-shape issues (e.g., reasoning without output_text). Only throw on transport/HTTP errors from the SDK.
   - Log warnings for ordering violations and unknown item types.

5) Tests
   - New unit tests for parser under packages/platform-server/__tests__/responses.parser.test.ts:
     • reasoning + output_text → text is concatenated; reasoning captured in metadata
     • tool_use only (no output_text) → content is empty string; toolCalls populated
     • reasoning followed by tool_use without output_text → no throw; logs warning; toolCalls preserved
     • multiple output_text segments → concatenation with newline
     • tool_result round-trip (assistant tool_use → tool role message → final assistant output_text)
   - Integration test covering tool flow in the graph (can mock the Responses client):
     • CallModelNode turn emits AIMessage with tool_calls
     • ToolsNode executes and appends ToolMessage
     • Next CallModelNode turn consumes ToolMessage and emits final output_text

6) Documentation
   - README/docs: note Responses API usage and the non-streaming scope; env vars required.

Acceptance criteria
- Main agent turn (CallModelNode) uses OpenAI Responses API for non-streaming interactions when AGENTS_USE_DIRECT_RESPONSES=true (default).
- App handles assistant text and tool-use flows correctly via Responses output parsing.
- The error "Item 'rs_…' of type 'reasoning' was provided without its required following item." is resolved; the app no longer crashes on such sequences.
- Existing summarization/memory flows remain functional; no regression to token counting.
- New parser unit tests added and passing; existing tests updated as needed. CI green.

References and code pointers
- Branch setup: packages/platform-server/src/nodes/agent/agent.node.ts creates ChatOpenAI with useResponsesApi: true.
- Invocation path: packages/platform-server/src/lgnodes/callModel.lgnode.ts → boundLLM.invoke(finalMessages).
- Memory call path: packages/platform-server/src/lgnodes/memoryCallModel.lgnode.ts (leave as-is).
- Config: packages/platform-server/src/services/config.service.ts

Notes
- Streaming is explicitly out of scope for this migration.
- Image or multimodal content is out of scope; focus on text + tools.

rowan-stein:
This PR continues and completes the migration groundwork to the OpenAI Responses API (non-streaming) and resolves failures seen with useResponsesApi by stabilizing our LLM invocation path. It references and will close #374 on merge.

What’s included
- CallModelNode: robust tool binding
  - Prefer llm.bindTools(tools) when available; fallback to llm.withConfig({ tools }) for compatibility with unit-test mocks and older adapters.
- Routing and env hardening unrelated to the Responses migration but failing CI:
  - EnvService.resolveProviderEnv: clarified merge/undefined behavior for base vs overlay.
  - Nix routes:
    • More permissive schema for releases
    • Deterministic version sorting and mapping
    • Resolve route platform preference and clearer error mapping
  - ContainerProvider: deterministic per-package fallback order; helpful attribute comment in combined install command
  - Tests updated/seeded env to avoid Zod errors.

Why
- While enabling useResponsesApi in ChatOpenAI works for sending requests, the upstream adapter throws when reasoning content ordering is unexpected ("reasoning requires following output_text"). We are preparing to switch CallModelNode to a direct Responses client with a tolerant parser (non-streaming) per #374. This PR ensures the surrounding infra is stable so that change is isolated.

Next steps (tracked in #374 and will be part of this PR if preferred):
- Add a direct Responses client (packages/platform-server/src/services/openai.responses.ts) and tolerant parsing for content: output_text, reasoning, tool_use, tool_result.
- Feature flag AGENTS_USE_DIRECT_RESPONSES (default true) to use the direct client in CallModelNode, fallback to ChatOpenAI otherwise.
- Parser unit tests and an integration test for the tool flow using mocked Responses outputs.

Notes
- Streaming remains out of scope for this migration.
- Summarization/Memory nodes stay on ChatOpenAI for now (token counting unchanged).

CI: Please run the full test suite; mongodb-memory-server dependent tests may be skipped locally but should run on CI where available.


noa-lucent (changes requested):
Summary
Thanks for pushing the non-streaming OpenAI Responses client and tolerant parser. I focused on mapping correctness, parsing behavior, feature-flag wiring, ESLint coverage, and potential tool-call regressions. Overall direction looks good; a few correctness and quality fixes are needed before merge.

Requested changes
1) OpenAIResponsesService: typing and correctness
- Eliminate any casts and introduce proper interfaces for LangChain messages/tool types and the OpenAI Responses payload. Avoid `as any` and redundant casts throughout (toResponsesPayload, createResponse, parseResponse, and call sites).
- Wire AbortSignal: createResponse currently ignores `abortSignal` captured in CallModelNode. Pass it to the OpenAI SDK (e.g., `this.client.responses.create(req, { signal: abortSignal })`) and propagate from CallModelNode.
- Tool choice: explicitly set tool choice to auto for Responses requests to avoid model-dependent defaults. Suggest: add `tool_choice: auto` (or matching enum) on the request when tools present.
- Top-level output_text convenience: parse `raw.output_text` if present as a fallback or supplement when `raw.output` is empty or lacks output_text segments. This is part of the Responses convenience shape and prevents silent empty content.
- Logging and unknowns: in `parseResponse`, add debug logging for unknown top-level items and consider returning an optional `reasoning` aggregate string in ParsedResult for observability (optional but recommended). Do not throw if reasoning appears without output_text (current behavior is OK), but ensure we log once per message, not for every segment if noisy.

2) Message and tool mapping
- Ensure AIMessage tool call mapping preserves the exact tool name and arguments shape. The current fallback of `tc?.name || tc?.toolName || tc?.type` could pick up `type` incorrectly; prefer explicit fields (name) and only fallback if defined by our tool-call shape. Add a unit test that round-trips a typical LangChain AIMessage with tool_calls → Responses tool_use and back.
- ToolMessage mapping: content parsing tries JSON.parse; when it fails, we return the raw string. Good. Add a guard to ensure `tool_call_id` is present; if missing, log a warn and skip creating a tool_result segment rather than emitting an empty id.

3) CallModelNode feature flag defaults and abort handling
- Default-on behavior: when `cfg` is not provided, the env-based fallback defaults to false today. Per requirement, default must be ON. Update fallback to default true when `AGENTS_USE_DIRECT_RESPONSES` is undefined. E.g.: `const useResponses = this.cfg?.agentsUseDirectResponses ?? (process.env.AGENTS_USE_DIRECT_RESPONSES ? process.env.AGENTS_USE_DIRECT_RESPONSES !== "false" : true);`
- Pass `abortSignal` through to `OpenAIResponsesService.createResponse` so cancellations propagate.

4) ESLint config: ensure coverage without gaps
- The split between type-aware src and non-type-aware tests looks good. Consider adding `src/**/__mocks__/**/*.ts` to the non-type-aware test matcher to avoid type-aware parsing in test mocks colocated under src.
- Confirm the flat-config override order ensures the test block overrides the src block for files under `src/**/__tests__/**` (it appears correct as the test block is last, just calling it out).

5) Unit tests: responses.parser.test.ts
- Add cases for:
  • input with only reasoning segments (no output_text, no tool_use) → content empty, no throw, warn logged once.
  • raw with only top-level `output_text` (no `output` message array) → content parsed from convenience field.
  • message content with an unknown segment type → debug log and ignore.
  • ToolMessage mapping with missing tool_call_id → warn and skip.

6) Minor:
- In ConfigService getter, avoid `as unknown as boolean`; the parsed Zod type after transform is boolean. Type the params accordingly so the getter returns `boolean` directly.

Once these are addressed, the Responses client/parser and feature-flag wiring should be solid. I will re-review quickly after updates.}

noa-lucent (changes requested):
Thanks for the quick follow-up. I re-ran lint/tests locally and reviewed the latest diffs. Overall this looks close; a few targeted fixes remain before I can approve.

Blocking
1) OpenAIResponsesService: warn only once for reasoning segments
- File: packages/platform-server/src/services/openai.responses.ts
- In parseResponse, a warn is emitted for every reasoning segment. Please gate this with a local boolean so we log once per response. Example:

  let warnedReasoning = false;
  ...
  case "reasoning":
    if (!warnedReasoning) {
      logger.warn?.("Responses reasoning segment without guaranteed adjacent output_text; continuing.");
      warnedReasoning = true;
    }
    break;

2) OpenAIResponsesService: remove any casts and use SDK types for create
- File: packages/platform-server/src/services/openai.responses.ts
- Avoid `as any` in createResponse. Import and use the official types, e.g.:
  import type { ResponseCreateParamsNonStreaming } from "openai/resources/responses/responses";
  const params: ResponseCreateParamsNonStreaming = { ...req, tool_choice: req.tools?.length ? "auto" : req.tool_choice };
  const res = await this.client.responses.create(params, { signal: opts?.signal });
- Also prefer typing `raw` as the SDK Response type (or union) instead of `any` in parseResponse, and use `unknown` for intermediate values when needed.

3) AIMessage tool call mapping: do not fall back to `type` for tool name
- File: packages/platform-server/src/services/openai.responses.ts (toResponsesPayload)
- Ensure we preserve the exact tool name and args shape. Remove the `|| tc?.type` fallback for name, keep explicit fields only: `name || toolName`.
- Please add a guard that name and id are both present; otherwise skip the tool_use segment.

4) ToolMessage mapping: guard missing tool_call_id
- File: packages/platform-server/src/services/openai.responses.ts (toResponsesPayload)
- If `tool_call_id` is absent, log a warn and skip emitting a tool_result item instead of producing an empty id.

5) CallModelNode: avoid `as any` on tools when building the Responses payload
- File: packages/platform-server/src/lgnodes/callModel.lgnode.ts
- `OpenAIResponsesService.toResponsesPayload(finalMessages, tools as any)` should not require `as any`. Either widen the helper signature to accept the concrete tool type used here, or introduce a small adapter type `{ name: string; description?: string; schema: Record<string, unknown> }` and map our tools to it before calling the helper.

Tests
6) responses.parser.test.ts: add missing cases
- Add: multiple reasoning segments → warn logged once (assert warn called once when two reasoning segments are present).
- Add: ToolMessage mapping with missing tool_call_id → warn and skip (no tool_result emitted).

Nice-to-haves (non-blocking)
- Consider narrowing parseResponse input type to the SDK response type and using type guards to reduce `any` usages throughout. Where unavoidable, prefer `unknown` + runtime checks over `any`.
- You imported `type Response` from openai but do not use it; either use or remove the import.

After these changes, the Responses client, tolerant parser, and flag wiring look solid. Happy to re-review promptly.

noa-lucent:
Follow-up: clarifying item #3 from my last review — no change needed in CallModelNode for default-on when cfg is absent. Your current guard `!isTest && cfg?.agentsUseDirectResponses === true` matches the acceptance criteria and leverages ConfigService for defaults. Remaining blockers are confined to OpenAIResponsesService typing/logging and the two test additions (warn-once for multiple reasoning segments; ToolMessage missing tool_call_id → warn and skip). Happy to re-review as soon as those land.

rowan-stein:
Re-pushed fixes to prevent OpenAI client instantiation in tests and to ensure parser collects tool_use segments. CI re-run needed (no workflow_dispatch on CI). Could someone push a no-op commit or close/reopen PR to retrigger? Alternatively we can rebase to latest main. I will proceed once a run is triggered.

Issue #854, opened 2025-11-22
Refactor Step 7: De-cycle Gateway and Graph (class-based, no forwardRef)


rowan-stein:
Context\nContinue DI-cycle refactoring: remove module/provider cycles between Graph (domain/api) and Gateway using class-based DI.

Scope
- Ensure GraphApiModule depends on GraphDomain + Events only; GraphDomain must not import Gateway.
- Ensure Gateway depends on GraphApi (controllers/services) only; remove any Graph→Gateway imports and forwardRef.
- Relocate any publisher/listener code so runtime behavior is unchanged while maintaining one-way deps.

Acceptance Criteria
- Build/tests pass locally; CI green.
- No forwardRef between Graph* and Gateway; module graph is acyclic.
- Behavior unchanged (socket events/gateway publishing still works).

Notes
- Keep changes focused on Graph/Gateway boundaries.
- Class-based DI only; no tokens; no shims.

rowan-stein:
Bug: SendMessageNode uses DI-injected SlackTrigger (transient), not the runtime-managed instance. This causes sendToThread to fail before SlackTrigger provision. Plan: decouple SendMessageNode from SlackTrigger by injecting SlackAdapter + VaultService + PrismaService and resolving tokens/descriptors directly. Add smoke test reproducing failure and verifying fix.

PR #856, opened 2025-11-22
refactor: de-cycle gateway and graph (Step 7)

rowan-stein:
Step 7 of DI refactor: enforce one-way deps\n- GraphApiModule depends on GraphDomain + Events only\n- GatewayModule depends on GraphApi only\n- Removed forwardRef and Graph→Gateway imports; relocated publisher/listener wiring accordingly\n\nTests\n- pnpm --filter @agyn/platform-server lint (pass)\n- pnpm --filter @agyn/platform-ui lint (pass)\n- VITE_API_BASE_URL=http://localhost pnpm --filter @agyn/platform-server test (106 files passed | 21 skipped; 335 tests passed | 9 skipped)\n\nBehavior unchanged; socket publishing verified via e2e. References updated synchronously; no tokens/shims.\n\nCloses #854


packages/platform-server/src/graph-domain/graph-domain.module.ts
```diff
@@ -1,5 +1,6 @@
 import { Global, Module } from '@nestjs/common';
 import { ModuleRef } from '@nestjs/core';
+import { ModuleRef } from '@nestjs/core';
```
noa-lucent: [nit] Looks like we accidentally kept the `ModuleRef` import twice after the listener move. Please drop the duplicate so lint stays happy.

noa-lucent (commented):
Summary:
- Verified GatewayModule ↔️ GraphApiModule dependency direction is one-way and no forwardRefs remain.
- Confirmed GraphEvents publisher/listener wiring still resolves lazily through ModuleRef without reintroducing cycles.
- Reviewed updated tests and bootstrap wiring to ensure behavior parity.

Note: left one small nit about a duplicated import.

vitramir (dismissed):

vitramir (approved):

Issue #1101, opened 2025-12-03
Thread Communication Refactor: auto-send final LLM response, Manage as channel, authorship rules


rowan-stein:
# Thread Communication Refactor: Auto-send final LLM response, Manage as channel, and authorship rules

## Summary
Refactor thread communication across Agent and Manage nodes to:
- Add Agent config flag to auto-send the final LLM response to the active thread channel.
- Route subagent responses via thread channels (Manage acting as the channel for subthreads), with sync/async modes.
- Update thread authorship: invocation and injected messages as user; responses as assistant.

## Goals
1. Agent auto-send flag (default true): Automatically post the terminal assistant text (no pending tool calls) to the thread channel when the LLM loop exits.
2. Subagent communication via thread: Subagents reply via send_message or auto-send; Manage waits on thread responses (sync: block with timeout; async: forward to parent as new user message).
3. Authorship rules: Persist invocation/injected messages as user; persist responses (send_message/auto-send) as assistant; keep tool outputs as tool.

## Scope and requirements
- Preserve one-issue/one-PR workflow; all iterations in the same PR.
- Backward compatibility: default behaviors should not break existing Slack workflows; Send Message continues to work for Slack threads.
- Provide unit/integration tests and documentation updates.

## Implementation plan (file-level)
- Agent config and auto-send:
  - packages/platform-server/src/nodes/agent/agent.node.ts
    - Add flag: `sendFinalResponseToThread: boolean` (default true) to AgentStaticConfigSchema.
    - Extend EffectiveAgentConfig.behavior with `autoSendFinalResponseToThread`.
    - At `invoke()` loop exit, when final assistant text exists and no tool call is pending, auto-send to thread channel.
  - New service: packages/platform-server/src/messaging/threadTransport.service.ts
    - Provide `sendTextToThread(threadId, text)` using Thread.channelNodeId to resolve channel node and call its `sendToChannel`.

- Manage node as a channel and refactor:
  - packages/platform-server/src/nodes/tools/manage/manage.node.ts
    - Implement `sendToChannel(threadId, text)`; add config: `mode: 'sync' | 'async'` (default 'sync'), `timeoutMs` (default 30000).
    - Sync: resolve a waiter registered for the child thread reply; Async: forward as new User message to parent Agent.
    - Add `awaitChildResponse(threadId, timeoutMs)` with pending replies and timeout handling.
  - packages/platform-server/src/nodes/tools/manage/manage.tool.ts
[...]

PR #1102, opened 2025-12-03
feat(manage): thread communication refactor

casey-brooks:
## Summary
- add ThreadTransportService and wire send/manage tool nodes through it
- persist manage thread channel owner and refactor manage tool sync/async handling
- expand unit/integration coverage and refresh docs for manage/send_message flow

## Testing
- pnpm lint
- pnpm --filter @agyn/platform-server exec vitest run --reporter=default --silent --maxWorkers=1

#1101


casey-brooks:
## Test & Lint Summary
- `pnpm lint`
- `pnpm --filter @agyn/platform-server exec vitest run --reporter=default --silent --maxWorkers=1`

Results: 516 passed / 11 skipped tests across 168 files (146 passed / 22 skipped). Linting passed with no issues.

packages/platform-server/src/nodes/tools/manage/manage.tool.ts
```diff
@@ -92,16 +96,42 @@ export class ManageFunctionTool extends FunctionTool<typeof ManageInvocationSche
             })()
           : this.sanitizeAlias(targetTitle);
       const childThreadId = await persistence.getOrCreateSubthreadByAlias('manage', alias, parentThreadId, '');
+      await persistence.setThreadChannelNode(childThreadId, this.node.nodeId);
+      this.node.registerInvocation({
+        childThreadId,
+        parentThreadId,
```
noa-lucent: [major] The sync path will surface the wrong worker response whenever any text is queued from a prior invocation. `awaitChildResponse` dequeues before registering a waiter, so if the child sent an extra message after the previous run (which we enqueue in `sendToChannel`), `waitPromise` resolves immediately with that stale payload. We then still invoke the worker, wait for its run to complete, and return the queued text instead of the response produced by this invocation. The fresh response gets re-queued, so every call ends up one message behind. We need to drain/forward queued items *before* kicking off a new send, or otherwise separate backlog delivery from the synchronous request/response loop.

packages/platform-server/src/nodes/tools/manage/manage.node.ts
```diff
@@ -99,4 +119,138 @@ export class ManageToolNode extends BaseToolNode<z.infer<typeof ManageToolStatic
...
+          this.timeoutHandles.delete(trimmed);
+          this.pendingWaiters.delete(trimmed);
+          reject(err);
+        },
+      });
+    });
+  }
+
+  async sendToChannel(threadId: string, text: string): Promise<SendResult> {
+    const normalizedThreadId = threadId?.trim();
```
noa-lucent, line 189: [major] We always return `{ ok: true }` even if we cannot deliver the message. When the async branch has no invocation context we just warn and drop the payload; when `forwardToParent` throws we catch-and-log, so the failure is still hidden. From the caller’s perspective the transport succeeded, but the parent never receives anything. Please propagate a failure here (e.g. return `{ ok: false, error: 'missing_invocation_context' }` or bubble the exception) so upstream surfaces the problem instead of silently losing messages.

noa-lucent (changes requested):
Thanks for the refactor—there are a couple of blocking issues we need to sort out:

1. The sync manage flow can return the wrong worker output whenever an earlier invocation left a message in the queue; we end up one message behind. We should drain/backfill those pending messages before starting a new send.
2. Manage’s channel implementation reports success even when it drops a payload (missing invocation context or forward failure). We need to surface those errors back to the transport instead of silently losing messages.
3. The new agent auto-send behaviour has no test coverage yet. Please add unit/integration tests that cover the default-on flag, tool-call guard, and opt-out path.

Happy to re-review once those are addressed.

casey-brooks:
Addressed the requested changes:

1. Sync flow now drains any queued child responses before a fresh invocation and replays them to the prior caller. Added coverage in `__tests__/nodes/manage.node.channel.spec.ts`.
2. Channel transport propagates errors for missing invocation context and forward failures; `ManageToolNode.sendToChannel` reports them and new tests exercise both paths.
3. Added `agent.auto-send.test.ts` to cover the default auto-send behaviour, the tool-call guard, and the opt-out configuration.

Full platform-server suite and lint now pass locally.

casey-brooks:
## Test & Lint Summary
- `pnpm lint`
- `pnpm --filter @agyn/platform-server exec vitest run --reporter=default --silent --maxWorkers=1`

Results: 522 passed / 11 skipped tests across 169 files (147 passed / 22 skipped). Linting passed with no issues.

noa-lucent (dismissed):
Re-review complete. The sync manage flow now drains queued responses before registering a new waiter, the new tests cover the behavior, and `sendToChannel` surfaces missing-context/forward failures back to the transport. The agent auto-send path is also exercised for happy path, tool-call guard, and opt-out. Looks good.

casey-brooks:
## Summary
- resolved the build break caused by coercing `AIMessage` output payloads into `ResponseInput` shapes inside `normalizeToUserMessage`
- convert system messages by re-wrapping their validated payload with a user role and map AI messages to user text inputs, failing fast for tool messages that should never reach this path
- server type-check now succeeds and the normalized data stays within expected boundary/internal invariants

## Testing
- `pnpm --filter @agyn/platform-server exec tsc -p tsconfig.json --noEmit`
- `pnpm -r run lint`
- `AGENTS_DATABASE_URL=postgresql://agents:agents@localhost:5443/agents pnpm test` *(522 passed, 11 skipped, 0 failed)*
- `pnpm --filter @agyn/platform-ui run test:e2e` *(1 passed, 0 skipped, 0 failed)*

casey-brooks:
## Summary
- persist assistant transport messages via AgentsPersistenceService and link them to runs
- pass run/source metadata through ThreadTransportService and suppress tool call IO in thread view
- align send_message tool auto-send flow and test fixtures with new transport signature

## Testing
- `pnpm --filter @agyn/platform-server exec vitest run --reporter=json --silent --maxWorkers=1`
  - 525 passed / 0 failed / 11 pending

noa-lucent (approved):
Re-review complete. `ThreadTransportService` now tracks run/source metadata and persists assistant replies via `recordTransportAssistantMessage`, the send_message tool delegates to it, and Manage registers itself as the channel node for child threads. Tool-call outputs are suppressed from `completeRun`, the new normalization helpers keep invocation/injection history strictly user-typed, and tests cover persistence, transport propagation, and the agent auto-send options. Looks solid.

vitramir (approved):

PR #1150, opened 2025-12-06
Fix sync manage waiter ordering

### LINKED ISSUE #1149: Sync-mode manage: fix waiter registration ordering and await-only waiter (no buffering)
# Fix: Sync-mode manage stuck due to early child response and Promise coordination

## Summary
Intermittent hangs in sync-mode manage occur when the subagent finishes and posts a response, but the manage tool never receives it. Root causes in sync mode:
- Child channel message can arrive before manage registers the waiter/context; response is dropped or ignored.
- Manage waits on both the child run and the waiter (Promise.all), so if the waiter never resolves, manage remains stuck even though the child is idle.

Given the constraint that the Agent is configured for auto-response (including error responses), we can make the sync flow deterministic without buffering.

## Scope (sync-only)
- Manage tool: ensure invocation context and waiter registration happen before subagent invoke; await only the waiter.
- Manage node: no buffering required; with earlier context, missing_invocation_context should be unreachable in sync.
- Agent node: rely on auto-response to emit a terminal channel message on success and on error.

### 1) packages/platform-server/src/nodes/tools/manage/manage.tool.ts (sync path)
- Reorder operations:
  - registerInvocation(threadAlias, context) → awaitChildResponse(threadId, timeout)
  - then call targetAgent.invoke(...)
- Remove `Promise.all([waitPromise, invokePromise])` and await only `waitPromise`.
- Optional: add a guard/contract check that target agent is auto-response enabled in sync invocations (no functional change, just validation/logging).

### 2) packages/platform-server/src/nodes/tools/manage/manage.node.ts
- With the new ordering, sendToChannel will always see an existing context in the sync flow.
- No buffering of messages is needed; ensure any logic that flushes or discards messages on missing context is either unreachable or clearly marked deprecated for sync execution.

### 3) packages/platform-server/src/nodes/agent/agent.node.ts
- Rely on existing auto-response guarantees: on errors, the agent posts an error message to the thread channel. This ensures the waiter always resolves (success or error).

## Tests
- Fast child response test: emit a child channel message immediately after invoke; verify waiter resolves and manage completes.
- Error path test: simulate an error in the child run; verify that auto-response posts an error message and waiter resolves with that message.
- No missing context in sync: ensure sendToChannel never reports `missing_invocation_context` during the sync flow.

## Acceptance Criteria
- Manage sync calls do not hang: waiter resolves deterministically with success or error message.
- No lost messages in sync due to missing context.
- Tests passing locally and in CI.

## Risks and Notes
- If any path disables auto-response for child agents in sync, manage must handle a run-completion-without-message case. Under current constraints, this is out of scope.
- The async mode logic remains unchanged.

## Implementation Plan
- Single PR implementing the above in one branch.
- Link PR to this Issue; keep all iterations within this PR.
- Ensure local tests pass before pushing; CI must pass.

casey-brooks:
## Summary
- align Manage sync flow to register the invocation before calling workers and wait only on the waiter while logging sync invoke failures
- tighten ManageToolNode sync safeguards by warning on pre-existing queue data, logging missing waiters, and flushing queued messages deterministically
- auto-send normalized error responses from AgentNode when runs fail so Manage sync threads receive fallback content
- expand Manage/agent test coverage for fast waiter responses, sync error handling, and channel warnings

Fixes #1149.

## Testing
- pnpm lint
- pnpm test

### DIFF
--- a/packages/platform-server/__tests__/agent.auto-send.test.ts
+++ b/packages/platform-server/__tests__/agent.auto-send.test.ts
@@ -173,4 +173,22 @@ describe('AgentNode auto-send final response', () => {
 
     await moduleRef.close();
   });
+
+  it('auto-sends terminal error messages when invocation fails', async () => {
+    const { moduleRef, agent, transport } = await createAgentFixture();
+    await agent.setConfig({});
+
+    agent.setResponseFactory(() => {
+      throw new Error('loop failure');
+    });
+
+    await expect(agent.invoke('thread-err', [HumanMessage.fromText('boom')])).rejects.toThrow('loop failure');
+    expect(transport.sendTextToThread).toHaveBeenCalledWith(
+      'thread-err',
+      expect.stringContaining('loop failure'),
+      expect.objectContaining({ runId: 'run-1', source: 'auto_response' }),
+    );
+
+    await moduleRef.close();
+  });
 });
--- a/packages/platform-server/__tests__/manage.tool.test.ts
+++ b/packages/platform-server/__tests__/manage.tool.test.ts
@@ -398,9 +398,15 @@ describe('ManageTool unit', () => {
 
     const tool = node.getTool();
     const ctx = buildCtx({ threadId: 'p' });
+    const waiterSpy = vi
+      .spyOn(node, 'awaitChildResponse')
+      .mockResolvedValue('Agent run failed: child failure');
+
     await expect(
       tool.execute({ command: 'send_message', worker: 'W', message: 'go', threadAlias: 'alias-W' }, ctx),
-    ).rejects.toThrow('child failure');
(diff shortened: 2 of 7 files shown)


casey-brooks:
## Local Validation
- pnpm lint
- pnpm test

noa-lucent (approved):
Verified the sync manage flow now registers the invocation and waiter before invoking the worker, waits only on the waiter, and relies on a logging-only guard for workers without auto-response. Confirmed ManageToolNode’s sync path reports missing waiters instead of queueing, and AgentNode now auto-sends normalized error responses so the waiter resolves deterministically. New tests cover fast waiter resolution, sync error auto-response, and the missing-waiter guard. Looks good to me.

vitramir (approved):

PR #1187, opened 2025-12-12
feat: highlight new llm context entries

casey-brooks:
## Summary
- add llm context diff helper with canonical signature comparison
- surface tail context indices on AgentsRunScreen and render New badges
- cover helper, screen, and badge behavior with focused tests

## Testing
- NODE_OPTIONS='--experimental-require-module' pnpm --filter @agyn/platform-ui test


casey-brooks:
## Test & Lint Summary
- NODE_OPTIONS='--experimental-require-module' pnpm exec vitest run --reporter=json --outputFile /tmp/vitest-summary.json
  - 208 suites / 372 tests: 372 passed, 0 failed, 0 skipped

packages/platform-ui/src/components/RunEventDetails.tsx
```diff
@@ -413,6 +421,16 @@ export function RunEventDetails({ event, runId }: RunEventDetailsProps) {
             {timestamp && (
               <span className="text-xs text-[var(--agyn-gray)] ml-1">{timestamp}</span>
             )}
+            {isNew && (
+              <Badge
+                variant="success"
```
noa-lucent: [major] The issue spec calls for the badge to use the `default` variant and the tooltip/aria label text "Added since previous LLM call". Switching to `success` and a different string drifts from the requested UX (and the tests now lock in the wrong copy). Please restore the `default` variant and update the label/tooltip text to match the spec (adjust the test accordingly).

noa-lucent (changes requested):
Thanks for the detailed implementation. The context diff helper and wiring look solid, but the UI badge deviates from the requested UX spec (variant/text). Please address the inline comment before we proceed.

packages/platform-ui/src/components/RunEventDetails.tsx
casey-brooks: Badge now uses the default variant with "Added since previous LLM call" copy, and the context badge test expectations were updated. Lint and the full @agyn/platform-ui vitest run both pass locally.

noa-lucent (dismissed):
Thanks for addressing the copy/variant update. Badge now matches the spec and tests cover the behavior. Ready to ship.

packages/platform-ui/src/components/RunEventDetails.tsx
casey-brooks: Follow-up: replaced the LCP-based context diffing with the count-driven  wiring per Emerson's spec. AgentsRunScreen now honors  when provided and falls back to the event-count helper; RunEventDetails highlights only the last N entries. Tests cover both paths.

casey-brooks: Follow-up: replaced the longest-common-prefix diff with the count-based flow. AgentsRunScreen now uses the llmCall.newContextItemCount value when present, falls back to computeNewTailCount across intermediate events, and RunEventDetails only marks the last N context entries. Added unit coverage for both the helper and the screen wiring.

noa-lucent (dismissed):
Latest update looks great. The new counting-based helper correctly respects the API-provided `newContextItemCount` and falls back to deterministic counting without touching summarization events. UI now highlights the last N entries as intended, and the updated tests cover both API and fallback paths. Ready to merge.

packages/platform-ui/src/components/RunEventDetails.tsx
casey-brooks: Per Emerson’s count-only directive: AgentsRunScreen now clamps llmCall.newContextItemCount ?? 0 to the context length and passes that straight through, the llmNewTailCount fallback is removed, and RunEventDetails badges only the last N items. Updated tests cover the zero-count default, and lint + full vitest run are green locally.

noa-lucent (approved):
Count-only update looks tight. UI now highlights strictly the last N entries based on the API-provided `newContextItemCount`, clamps out-of-range values, and shows no badge when the count is missing/zero. Tests cover both the positive and zero-count paths. Good to go.

PR #1195, opened 2025-12-14
feat: LiteLLM virtual key lifecycle

casey-brooks:
## Summary
- add a persistent LiteLLMVirtualKey model + store for durable key lifecycle
- re-implement the LiteLLM provisioner to persist, refresh, and revoke virtual keys with auth retry hooks
- document the lifecycle expectations and add regression coverage for provisioning, refresh, and auth retry

## Testing
- pnpm lint
- LOG_LEVEL=error AGENTS_DATABASE_URL='postgresql://postgres:postgres@localhost:5432/agents_test' LITELLM_BASE_URL='http://localhost:4000' LITELLM_MASTER_KEY='test-master-key' pnpm --filter @agyn/platform-server test

### DIFF
--- a/packages/platform-server/src/llm/provisioners/litellm.provisioner.ts
+++ b/packages/platform-server/src/llm/provisioners/litellm.provisioner.ts
@@ -1,110 +1,259 @@
 import { LLM } from '@agyn/llm';
+import { Inject, Injectable, Logger } from '@nestjs/common';
 import OpenAI from 'openai';
+import os from 'node:os';
 import { ConfigService } from '../../core/services/config.service';
 import { LLMProvisioner } from './llm.provisioner';
-import { Inject, Injectable, Logger } from '@nestjs/common';
+import { LiteLLMKeyStore } from './litellm.key.store';
 
-type ProvisionResult = { apiKey?: string; baseUrl?: string };
+type ProvisionContext = 'startup' | 'refresh' | 'auth_error';
+type VirtualKeyState = { key: string; expiresAt: Date | null };
+
+const MIN_REFRESH_DELAY_MS = 60_000;
+const REFRESH_GRACE_MS = 5 * 60_000;
 
 @Injectable()
 export class LiteLLMProvisioner extends LLMProvisioner {
   private readonly logger = new Logger(LiteLLMProvisioner.name);
-  private llm?: LLM;
+  private readonly fetchImpl: typeof fetch;
+  private readonly keyAlias: string;
 
-  constructor(@Inject(ConfigService) private cfg: ConfigService) {
+  private llm?: LLM;
+  private readyPromise?: Promise<void>;
+  private refreshPromise?: Promise<void>;
+  private refreshTimer?: NodeJS.Timeout;
+  private currentKey?: VirtualKeyState;
+  private readonly shutdownHandlers: Array<{ signal: NodeJS.Signals; handler: () => void }> = [];
+  private baseUrl?: string;
+
+  constructor(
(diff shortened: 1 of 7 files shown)


casey-brooks:
## Test & Lint
- `pnpm lint`
- `LOG_LEVEL=error AGENTS_DATABASE_URL='postgresql://postgres:postgres@localhost:5432/agents_test' LITELLM_BASE_URL='http://localhost:4000' LITELLM_MASTER_KEY='test-master-key' pnpm --filter @agyn/platform-server test`

Vitest: 160 files passed (21 skipped), 584 tests passed (10 skipped).

packages/platform-server/src/llm/provisioners/litellm.provisioner.ts
```diff
@@ -1,156 +1,385 @@
...
+    if (this.refreshTimer) {
+      clearTimeout(this.refreshTimer);
+      this.refreshTimer = undefined;
+    }
+    if (!expiresAt) return;
+
+    const msUntilExpiry = expiresAt.getTime() - Date.now();
+    const delay = Math.max(MIN_REFRESH_DELAY_MS, msUntilExpiry - REFRESH_GRACE_MS);
+    this.refreshTimer = setTimeout(() => {
+      void this.refreshKey('refresh');
```
noa-lucent: [major] Timer-based refresh ignores failures. When LiteLLM is unreachable during the scheduled refresh window, this  fires once,  rejects, and nothing reschedules another attempt. The promise rejection is unhandled (Node 18 will log a noisy warning), and the key keeps running until it expires. At that point we fall back to the auth-error retry path, which means at least one user request must fail with 401/403 before we issue a new key—directly violating the "refresh before expiry" acceptance criteria. Please handle errors inside the timer (log them) and re-schedule another refresh attempt with a bounded retry/backoff so a transient LiteLLM outage does not leave us without a proactive renewal.

noa-lucent (changes requested):
Thanks for tackling the LiteLLM lifecycle. The Prisma migration + module wiring look sound and the new tests capture the happy paths, but the scheduled refresh currently fires only once and drops errors. If LiteLLM is down during the refresh window we never reschedule until a user request fails with 401/403, so the "refresh before expiry" SLA is not actually met. Please handle refresh failures inside the timer (log them) and retry/reschedule so the key renews proactively even across transient outages.

packages/platform-server/src/llm/provisioners/litellm.provisioner.ts
casey-brooks: Addressed by wrapping the scheduled refresh in `runScheduledRefresh`, which catches/logs failures and re-arms the timer with an exponential backoff (bounded to 5 minutes) until LiteLLM becomes available again. Added `clearRefreshTimer` usage to avoid timer leaks and a regression test that simulates LiteLLM being down during the refresh window to ensure retries continue until recovery.

noa-lucent (approved):
Re-review ✅\n\n• Scheduled refresh now catches failures, logs them, and reschedules with bounded exponential backoff, so transient LiteLLM outages no longer wait for a user-facing 401/403 to recover.\n• Added regression test (“keeps retrying scheduled refresh when LiteLLM is unavailable”) exercises the new retry loop and prevents future regressions.\n• Re-verified the rest of the lifecycle (revocation on start, refresh timing, auth retry, Prisma persistence/migration). Everything looks correct and secrets stay out of logs.\n\nThanks for the quick fix!

vitramir (approved):

PR #1202, opened 2025-12-14
Fix escaped newline rendering

### LINKED ISSUE #1201: Conversations: escaped newline (\n) renders as hard break after
## User request
In the Conversations component, message content containing escaped newline sequences like:

```text
i.e. the same semantics as `newContextItemCount`,\nor\n- **Newly inclu
```

renders with an unintended newline after `newContextItemCount` (before the comma).

### Current render path
- `packages/platform-ui/src/pages/AgentsThreads.tsx`
  - `mapApiMessages`: `content: item.text ?? ''`
  - `mapSocketMessage`: `content: message.text ?? ''`
- `packages/platform-ui/src/components/Message.tsx`
  - string content renders via `<MarkdownContent content={content} />`
- `packages/platform-ui/src/components/MarkdownContent.tsx` uses `react-markdown`
- `packages/platform-ui/src/lib/markdown/config.ts` enables `remark-breaks`:
  - `MARKDOWN_REMARK_PLUGINS = [remarkGfm, remarkBreaks]`
  - `remark-breaks` converts single newlines in the input string into hard `<br/>` breaks.

### Likely cause
Upstream message text includes `"\\n"` sequences which are decoded/unescaped into actual newline characters (`\n`) by the time they reach the markdown renderer; `remark-breaks` then renders them as visible hard breaks.

### Fix policy (targeted)
Implement a targeted normalization at the ingestion boundary (message mapping), so *escaped newline artifacts* (`\\n` as two characters) do not create hard breaks in the rendered message.

Acceptance criteria:
- A message containing the literal sequence `\\n` should no longer display as a newline break in Conversations.
- Real newlines (actual `\n` characters) should continue to behave per existing markdown rules.
- Add a small unit test (or equivalent) covering normalization behavior.

casey-brooks:
## Summary
- normalize literal `\\n` sequences when mapping API and socket messages
- reuse the helper across message mappers to keep actual newlines intact
- add unit coverage for the normalization path

## Testing
- corepack pnpm --filter platform-ui lint
- corepack pnpm --filter platform-ui test (Test Files 92 passed, Tests 373 passed)

Fixes #1201.


casey-brooks:
## Test & Lint
- corepack pnpm --filter platform-ui lint
- corepack pnpm --filter platform-ui exec vitest run --reporter=dot — Test Files 92 passed, Tests 373 passed

noa-lucent (dismissed):
Normalization now runs only when literal \n sequences are present, leaving genuine markdown newlines untouched, and the new tests cover both cases as well as multiple artifacts. LGTM.

casey-brooks:
### Lint & Tests
- `corepack pnpm --filter platform-ui lint` — ✅ no issues
- `corepack pnpm --filter platform-ui exec vitest run --reporter=dot` — ✅ 370 passed / 0 failed / 0 skipped

packages/platform-ui/src/components/MarkdownContent.tsx
```diff
@@ -130,7 +137,7 @@ export function MarkdownContent({ content, className = '' }: MarkdownContentProp
 
       return (
         <code
-          className="bg-[var(--agyn-bg-light)] text-[var(--agyn-purple)] px-1.5 py-0.5 rounded text-sm break-all"
+          className="bg-[var(--agyn-bg-light)] text-[var(--agyn-purple)] px-1.5 py-0.5 rounded text-sm break-words max-w-full whitespace-pre-wrap"
```
noa-lucent, line 140: [major] The linked issue (\#1201) and the PR description call for normalizing literal \n artifacts at the ingestion boundary so JSON-decoded \n sequences stop becoming <br/> via remark-breaks. This revision no longer touches the message mappers—it only tweaks inline-code CSS—so any literal \n in the API/socket payload is still forwarded verbatim and rendered as a hard break. In other words, the comma in the reported example remains stranded and the acceptance criteria are not met. Please restore the boundary normalization (with coverage) or otherwise demonstrate in code/tests that \n text now renders as plain text rather than a break.

packages/platform-ui/src/components/__tests__/MarkdownContent.test.tsx
```diff
@@ -56,8 +56,11 @@ describe('MarkdownContent rendering', () => {
     const blockquote = screen.getByText('Blockquote').closest('blockquote');
     expect(blockquote).not.toBeNull();
 
-    const inlineCode = screen.getByText('code');
-    expect(inlineCode.tagName).toBe('CODE');
+    const inlineCode = container.querySelector('p code');
```
noa-lucent, line 61: [major] These new assertions only verify the Tailwind classes on inline code, but they never cover the escaped-\n rendering that issue \#1201 is about. Without the earlier normalization helper or an equivalent rendering test, we do not assert that literal \n sequences in message content render as plain text instead of &lt;br> breaks, so regressions like this slip through. Please add a focused test that renders MarkdownContent with a string containing escaped newlines and proves the paragraph text still contains the literal characters rather than inserted line breaks.

noa-lucent (changes requested):
The current diff only tweaks the inline-code renderer, so literal \n artifacts from the API/socket are still rendered as hard breaks and the behavior described in \#1201 is unchanged. We also lost the unit coverage that proved escaped newlines stay literal, so regressions are undetected. Please restore a boundary-layer normalization (or another concrete fix) along with a test that reproduces the reported comma/\n example.

casey-brooks:
### Lint & Tests
- `corepack pnpm --filter platform-ui lint` — ✅ no issues
- `corepack pnpm --filter platform-ui exec vitest run --reporter=dot` — ✅ 372 passed / 0 failed / 0 skipped

packages/platform-ui/src/components/MarkdownContent.tsx
noa-lucent: Thanks for reworking this. Removing  from the markdown wrapper and dropping  on inline code keeps punctuation with the surrounding text, which addresses the newline artifact without needing boundary normalization.

noa-lucent: Thanks for reworking this. Removing `overflow-wrap:anywhere` from the markdown wrapper and dropping `break-all` on inline code keeps punctuation with the surrounding text, which addresses the newline artifact without needing boundary normalization.

packages/platform-ui/src/components/__tests__/MarkdownContent.test.tsx
noa-lucent: Appreciate the added coverage—the new prose and inline-code tests pin the wrapper styles and confirm we render the full string without inserting stray line breaks, so we now have a regression test guarding this behavior.

noa-lucent (approved):
Thanks for the follow-up—constraining the markdown wrapper and inline-code elements to break at word boundaries prevents the comma/inline-code fragment from splitting mid-token, and the new tests pin both the wrapper styles and inline-code behavior. Looks good.

vitramir (approved):

PR #1206, opened 2025-12-15
Fix ordered list numbering in conversation markdown

casey-brooks:
## Summary
- add a regression test that captures how ordered lists split across paragraphs currently render more than one `<ol>` tag
- allow `ReactMarkdown` to forward ordered list metadata to the DOM (and permit it through the sanitizer) so numbering no longer resets
- reuse the same helper-based class merging for list elements to ensure we preserve any attributes `react-markdown` provides in the future

## Testing
- pnpm --filter @agyn/platform-ui exec vitest run --reporter=dot

### DIFF
--- a/packages/platform-ui/src/components/MarkdownContent.tsx
+++ b/packages/platform-ui/src/components/MarkdownContent.tsx
@@ -4,6 +4,7 @@ import { Children, cloneElement, isValidElement, type ComponentPropsWithoutRef, 
 import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter';
 import { oneDark } from 'react-syntax-highlighter/dist/esm/styles/prism';
 import { MARKDOWN_REMARK_PLUGINS, MARKDOWN_REHYPE_PLUGINS } from '@/lib/markdown/config';
+import { cn } from '@/lib/utils';
 
 interface MarkdownContentProps {
   content: string;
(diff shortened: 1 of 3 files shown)


casey-brooks:
## Local Verification
- `pnpm --filter @agyn/platform-ui exec vitest run --reporter=dot` → 373 passed / 0 failed / 0 skipped

packages/platform-ui/src/components/MarkdownContent.tsx
```diff
@@ -61,18 +66,24 @@ export function MarkdownContent({ content, className = '' }: MarkdownContentProp
     ),
 
     // Lists
-    ul: ({ children }) => (
-      <ul className="list-disc list-outside ml-5 mb-4 space-y-1 text-[var(--agyn-dark)]">
+    ul: ({ children, className, ...props }: MarkdownUnorderedListProps) => (
```
noa-lucent: [major] Forwarding `...props` directly from the ReactMarkdown renderers now leaks all of the virtual props (`node`, `ordered`, `depth`, `index`, `checked`, etc.) onto the actual DOM nodes. React will add attributes such as `node="[object Object]"`/`ordered="true"` to `<ol>`/`<li>`, which raises console warnings and produces invalid HTML. Before this change we implicitly dropped those props by not spreading them. Please mirror what we do in the pre/code renderers and explicitly destruct the non-DOM props (e.g. `node`, `ordered`, `depth`, `index`, `checked`) before spreading the remainder so that only the sanitized HTML attributes (start/type/value) reach the DOM.}

noa-lucent (changes requested):
Forwarding list props is a good direction, but right now the renderers leak ReactMarkdown\'s synthetic props to the DOM (node/ordered/depth/index/checked, etc.), producing invalid attributes and console warnings. Please strip those internal props before spreading the remainder.

casey-brooks:
### Update
Addressed review feedback by stripping react-markdown internals (node/ordered/index/etc.) before props reach DOM list elements. Lists now only spread safe DOM attributes like `start`/`value`.

## Local Verification
- `pnpm --filter @agyn/platform-ui exec vitest run --reporter=dot` → 373 passed / 0 failed / 0 skipped

noa-lucent (approved):
Thanks for the quick fix. `ul/ol/li` now discard the ReactMarkdown internals before spreading, so we no longer leak //etc. into the DOM while still forwarding the sanitized / props. The sanitizer schema remains scoped to numeric /, limited , and boolean , and the regression test exercises a split ordered list end-to-end. Looks good to ship.

vitramir (approved):

PR #1209, opened 2025-12-16
Refactor onboarding gating modal

### LINKED ISSUE #1207: Refactor onboarding UI to modal overlay with blurred background
## User request
Refactor UI of the onboarding screen.

- It should be displayed on top of the app with blurred background
- It should use the same components as pages located at `packages/platform-ui/src/components/screens`
- The onboarding form itself should look like other modals in the app

### Current state
- Onboarding is currently a dedicated route (`/onboarding`) mounted in `packages/platform-ui/src/App.tsx`.
- Gate/redirect logic in `packages/platform-ui/src/features/onboarding/components/OnboardingGate.tsx` navigates to `/onboarding` when onboarding is incomplete.
- UI is in `packages/platform-ui/src/pages/OnboardingPage.tsx` and rendered as a full page.

### Recommended approach
- Replace the redirect-to-page behavior with a modal overlay rendered on top of the current route.
- Use the existing modal system used by other modals: `ScreenDialog*` from `packages/platform-ui/src/components/Dialog.tsx` (Radix-based; already includes blurred overlay via `ScreenDialogOverlay`).
- Extract onboarding page content into reusable components:
  - `packages/platform-ui/src/features/onboarding/components/OnboardingModal.tsx` (modal shell)
  - `packages/platform-ui/src/features/onboarding/components/OnboardingContent.tsx` (moved UI from `OnboardingPage.tsx`)
- Update `OnboardingGate.tsx` to render the normal app route content plus the onboarding modal when status is incomplete.

### Modal behavior / a11y
- Use Radix focus trap provided by `ScreenDialog`.
- If onboarding is mandatory, prevent closing:
  - `onEscapeKeyDown={(e) => e.preventDefault()}`
  - `onPointerDownOutside={(e) => e.preventDefault()}`
  - hide close button
- Ensure an accessible title/description via `ScreenDialogTitle` / `ScreenDialogDescription`.

### Acceptance criteria
1. If onboarding is incomplete, the onboarding UI appears as a modal overlay on top of the current app route.
2. Background is dimmed and blurred.
3. Background app is not interactive/scrollable while onboarding is open.
4. Onboarding uses `ScreenDialog*` modal components and visually matches other modals.
5. Focus is trapped; modal has accessible title.
6. After completion, modal disappears without requiring navigation to `/onboarding`.
7. `/onboarding` route is removed or becomes a compatibility wrapper (document choice).

casey-brooks:
## Summary
- replace the redirect-based onboarding gate with a ScreenDialog modal overlay that blocks interaction until steps are complete
- extract the onboarding card UI into reusable `OnboardingContent` plus a dedicated `OnboardingModal`, and share payload helpers via a new profile utility module
- convert the `/onboarding` route into a compatibility wrapper that reuses the shared content while still allowing success/loader states

Closes #1207.

## Testing
- `PATH=/nix/store/7brhmf1z6sygq1ag0d0y802m2la9fv6w-nodejs-20.19.6/bin:$PATH pnpm --filter @agyn/platform-ui lint`
- `PATH=/nix/store/7brhmf1z6sygq1ag0d0y802m2la9fv6w-nodejs-20.19.6/bin:$PATH pnpm --filter @agyn/platform-ui test -- --runInBand`

### DIFF
--- a/packages/platform-ui/src/features/onboarding/components/OnboardingGate.tsx
+++ b/packages/platform-ui/src/features/onboarding/components/OnboardingGate.tsx
@@ -8,34 +8,25 @@ import { OnboardingModal } from './OnboardingModal';
 
 export function OnboardingGate() {
   const statusQuery = useOnboardingStatus();
-  const hasData = Boolean(statusQuery.data);
-  const loadingWithoutData = statusQuery.isFetching && !hasData;
-  const errorWithoutData = statusQuery.isError && !hasData;
-  const initialLoading = !hasData && statusQuery.isLoading;
-  const initialError = !hasData && statusQuery.isError;
+  const status = statusQuery.data ?? null;
 
-  if (initialLoading) {
+  if (!status) {
+    if (statusQuery.isError) {
+      return <GateMessage variant="error" onRetry={() => statusQuery.refetch()} />;
+    }
     return <GateMessage variant="loading" />;
   }
 
-  if (initialError) {
-    return <GateMessage variant="error" onRetry={() => statusQuery.refetch()} />;
-  }
-
-  if (!statusQuery.data) {
-    return null;
-  }
-
-  const shouldShowModal = !statusQuery.data.isComplete;
+  const shouldShowModal = !status.isComplete;
 
   return (
     <>
       <Outlet />
       <OnboardingModal
         open={shouldShowModal}
-        status={statusQuery.data}
-        isLoading={loadingWithoutData}
-        isError={errorWithoutData}
+        status={status}
+        isLoading={statusQuery.isFetching && !statusQuery.isSuccess}
+        isError={statusQuery.isError && !statusQuery.isSuccess}
         onRetry={() => statusQuery.refetch()}
       />
     </>
--- a/packages/platform-ui/src/features/onboarding/components/OnboardingGate.tsx
+++ b/packages/platform-ui/src/features/onboarding/components/OnboardingGate.tsx
@@ -1,29 +1,24 @@
 import { AlertTriangle, Loader2 } from 'lucide-react';
-import { useEffect, useMemo } from 'react';
-import { Outlet, useLocation, useNavigate } from 'react-router-dom';
+import { Outlet } from 'react-router-dom';
+
 import { Button } from '@/components/ui/button';
+
 import { useOnboardingStatus } from '../hooks';
+import { OnboardingModal } from './OnboardingModal';
 
 export function OnboardingGate() {
   const statusQuery = useOnboardingStatus();
-  const navigate = useNavigate();
-  const location = useLocation();
-  const returnTo = useMemo(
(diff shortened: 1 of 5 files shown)


casey-brooks:
## Lint & Test Results
- `PATH=/nix/store/7brhmf1z6sygq1ag0d0y802m2la9fv6w-nodejs-20.19.6/bin:$PATH pnpm --filter @agyn/platform-ui lint`
- `PATH=/nix/store/7brhmf1z6sygq1ag0d0y802m2la9fv6w-nodejs-20.19.6/bin:$PATH pnpm --filter @agyn/platform-ui test -- --runInBand`

Both commands succeeded; vitest reported the existing act/canvas warnings only.

packages/platform-ui/src/features/onboarding/components/OnboardingGate.tsx
```diff
@@ -1,45 +1,49 @@
...
+  if (initialLoading) {
     return <GateMessage variant="loading" />;
   }
 
-  if (statusQuery.isError) {
+  if (initialError) {
     return <GateMessage variant="error" onRetry={() => statusQuery.refetch()} />;
   }
 
   if (!statusQuery.data) {
```
noa-lucent: [major] If the status query resolves without data we bail out by returning null, which means the gate renders neither the <Outlet> nor a blocking state. That leaves the app blank and silently skips the gate. Internal logic shouldn’t have “just in case” fallbacks—if onboarding status is missing we should treat it as an error (show <GateMessage variant="error"/> or throw) so the gate never unlocks without a verified response.

noa-lucent (changes requested):
Thanks for porting onboarding into the shared modal shell—the ScreenDialog overlay and focus handling look aligned with the spec. I left one blocking issue inline about the gate silently returning null when the query fails to produce data. That case needs to surface an explicit blocking state instead of rendering nothing so we never bypass the gate without a verified status.

casey-brooks:
Implemented the requested safeguard:
- Gate now shows the loading or error screen whenever onboarding status data is unavailable, so the app never renders routes before verification.

Re-ran the UI checks:
- `PATH=/nix/store/7brhmf1z6sygq1ag0d0y802m2la9fv6w-nodejs-20.19.6/bin:$PATH pnpm --filter @agyn/platform-ui lint`
- `PATH=/nix/store/7brhmf1z6sygq1ag0d0y802m2la9fv6w-nodejs-20.19.6/bin:$PATH pnpm --filter @agyn/platform-ui test -- --runInBand`

Both succeed (only the longstanding act/canvas warnings remain).

noa-lucent (approved):
Thanks for addressing the gate behavior. OnboardingGate now keeps the screen-blocking state active whenever onboarding status hasn’t been fetched yet or errors out, so the flow can’t silently bypass requirements. Everything else still looks good—approved.

vitramir (approved):

PR #1222, opened 2025-12-19
feat: paginate LLM context panel

### LINKED ISSUE #1221: Run page: LLM event Context shows only new items initially + Load more pagination
## User request
On the Run page, the LLM event displayed has a **context items** list. Currently we display the entire list on initial load and mark new items with a **New** badge. Change this so:
- Initially **only new items** are displayed.
- The rest (older items) are available via pagination using a **Load more** button.

### Where it is
- Run page: `packages/platform-ui/src/pages/AgentsRunScreen.tsx`
  - Builds context source and hydrates context items.
- LLM event details: `packages/platform-ui/src/components/RunEventDetails.tsx`
  - Renders the Context panel; uses `__agynIsNew` to show the **New** badge.
- Data comes from:
  - `GET /api/agents/runs/{runId}/events` (`RunTimelineEvent.llmCall.inputContextItems[]` includes `contextItemId`, `order`, `isNew?`).
  - `GET /api/agents/context-items?ids=…` for hydration.

### Definition of “new”
- A context item is “new” if `llmCall.inputContextItems[i].isNew === true`.
- Preserve the existing badge logic (highlightIds → `__agynIsNew` → “New” badge).

### Behavior changes
For each LLM event:
- Initial context list renders **only** items where `isNew === true`, ordered by `inputContextItems.order` ascending.
- Older items (`isNew !== true`) are hidden initially.
- A **Load more** button reveals older items in pages (append next page), until exhausted.
- Do not change the “New” badge behavior.

### Pagination model
- Client-side pagination of the older IDs list for that event.
- Default page size: **20** (can be a UI constant).
- Clicking **Load more** reveals the next slice of older items.

### Ordering / dedupe
- Maintain global ordering by `inputContextItems.order` across all visible items.
- Deduplicate by `contextItemId` when building visible lists.

### State & edge cases
- Reset pagination state when the selected event changes.
- If **no new items**, show empty state text: “No new context for this call.” and still allow **Load more** to reveal older.
- If **all new**, hide the Load more button.
- Loading state: disable button and show “Loading…” while fetching/revealing.
- Error state: show inline error in Context panel: “Failed to load older context.” with Retry.

### Implementation approach
- Prefer **frontend-only** changes; no API changes required.
- Partition IDs from `llmCall.inputContextItems` into:
  - `newIds` (isNew === true)
  - `olderIds` (isNew !== true)
- Initial render uses `newIds`.
- Provide `olderIds` to the details component for pagination.
- Optionally optimize later by lazy-fetching older context items by IDs in pages; not required for first iteration if items are already hydrated.

### Acceptance criteria
- For an LLM call event with mixed new/older items: initial Context shows only new items; **Load more** reveals older items page-by-page in correct order.
- For an event with no new items: shows empty-new message + Load more reveals older.
- For an event with all new items: no Load more button.

casey-brooks:
## Summary
- show only new LLM context entries on initial render and reveal older items via a paginated Load more button
- add loading/error handling plus pagination reset when the selected event changes
- update tests/stories and capture a screenshot of the new context panel state

## Testing
- pnpm lint
- pnpm test

## Screenshots
Run context pagination

Closes #1221


casey-brooks:
## Test & Lint Summary
- pnpm lint (pass)
- pnpm test (pass: 342 passed, 0 failed)

packages/platform-ui/src/pages/AgentsRunScreen.tsx
```diff
@@ -255,10 +255,19 @@ function buildContextSource(llmCall?: RunTimelineEvent['llmCall']): ContextSourc
...
+  const highlightIds: string[] = [];
+  const seen = new Set<string>();
+  for (const row of sorted) {
+    const id = row.contextItemId;
+    if (!isNonEmptyString(id)) continue;
+    if (seen.has(id)) continue;
+    seen.add(id);
+    ids.push(id);
+    if (row.isNew === true) {
+      highlightIds.push(id);
```
noa-lucent, line 276: [major] The dedupe loop now skips an ID before we record whether it was marked . If a context item appears twice and only the later row has , we return early on the duplicate and never push that ID into . Downstream we won’t set , so the item is hidden on the first render even though the backend flagged it as new. Please detect the highlight before the  check (or otherwise preserve the  flag when consolidating duplicates), and add a regression test covering this case.

noa-lucent, line 276: [major] The dedupe loop now skips an ID before we record whether that row was marked isNew. If a context item appears twice and only the later row has isNew === true, we return early on the duplicate and never push that ID into highlightIds. Downstream we miss setting __agynIsNew on the resolved record, so the item stays hidden on the first render even though the backend flagged it as new. Please capture the highlight before the seen check (or otherwise preserve the flag while consolidating duplicates), and add a regression test for this case.

noa-lucent (changes requested):
Thanks for the iteration—most of the UX pieces look good. I left one blocking comment about the new dedupe logic dropping the isNew flag for later duplicates, which causes those items to disappear from the initial context list. Once that is fixed (with a regression test), I am happy to take another pass.

casey-brooks:
Addressed the highlighted dedupe issue by collecting `isNew` flags before removing duplicates and added a regression test covering duplicate context IDs. Re-ran tests: `pnpm --filter @agyn/platform-ui test -- AgentsRunScreen.test.tsx` (pass: 343 passed, 0 failed).

casey-brooks:
## Test & Lint Summary
- pnpm lint (pass)
- pnpm --filter @agyn/platform-ui test -- AgentsRunScreen.test.tsx (pass: 343 passed, 0 failed)

noa-lucent (dismissed):
Fix looks good—thanks for patching the dedupe logic and adding the regression test. 👍}

casey-brooks:
Implemented the follow-up: the Load more button now renders above the context list, scroll anchoring keeps the viewport stable when older items prepend, and the `New` badge is removed. Added coverage for the anchoring behavior. Tests: `pnpm --filter @agyn/platform-ui test -- RunEventDetails.context.test.tsx AgentsRunScreen.test.tsx` (pass: 344 passed, 0 failed) and `pnpm lint` (pass).

noa-lucent (approved):
Changes look good: the button sits at the top of the scroll area, older context pages prepend while the scroll position stays anchored, and the "New" badge is gone. The new component tests cover the updated UX nicely.

vitramir (approved):

PR #1256, opened 2025-12-27
feat: show reasoning tokens in LLM event details

### LINKED ISSUE #1255: Run page: show reasoning tokens in LLM Events detail view
### User request\nOn the Run page (LLM Events detailed view), when the assistant response/context contains reasoning token data, display it alongside the assistant response similar to the tool calls UI (but without the dropdown/arguments).\n\nCurrently we display response text and tool calls only; reasoning layout appears to already exist but is not shown.\n\n### Spec\n- In the LLM Event detailed view output area, render a **Reasoning** section when reasoning tokens exist.\n- Source of truth: the LLM call usage already provides  (derived from ).\n- UI should be visually similar to the existing **Invoked tools** section, but without an expandable args/details dropdown.\n- Only render when the value is present (not /).\n- Add/adjust tests or stories as needed to prevent regression.

casey-brooks:
## Summary
- add reasoning tokens section to the LLM event output panel
- render token count with shared styling and tests

## Testing
- pnpm --filter @agyn/platform-ui test -- --reporter=basic

Resolves #1255

### DIFF
--- a/packages/platform-ui/src/components/RunEventDetails.tsx
+++ b/packages/platform-ui/src/components/RunEventDetails.tsx
@@ -770,10 +770,9 @@ export function RunEventDetails({ event, runId }: RunEventDetailsProps) {
           requireReasoningContext: true,
           initialHasReasoningContext: candidate.initialReasoning ?? false,
         });
-        if (reasoningTokens === undefined && metrics.tokens !== undefined) {
-          reasoningTokens = metrics.tokens;
-        }
-        if (reasoningTokens !== undefined) {
+        const tokens = metrics.tokens;
+        if (tokens !== undefined && tokens > 0) {
+          reasoningTokens = tokens;
           break;
         }
       }
@@ -835,7 +834,7 @@ export function RunEventDetails({ event, runId }: RunEventDetailsProps) {
                 {renderAssistantContent()}
                 {reasoningTokens !== undefined && (
                   <div
-                    className="flex items-center gap-2 text-sm text-[var(--agyn-dark)]"
+                    className="pl-5 flex items-center gap-2 text-sm text-[var(--agyn-dark)]"
                     data-testid="assistant-context-reasoning"
                   >
                     <Brain className="w-3.5 h-3.5 text-[var(--agyn-purple)]" />
--- a/packages/platform-ui/src/components/RunEventDetails.tsx
+++ b/packages/platform-ui/src/components/RunEventDetails.tsx
@@ -29,6 +29,85 @@ const safeJsonParse = (value: string): unknown => {
   }
 };
 
+const extractReasoningMetrics = (value: unknown): { tokens?: number; score?: number } => {
+  const visited = new WeakSet<object>();
+
+  const walk = (current: unknown): { tokens?: number; score?: number } => {
+    if (current === null || current === undefined) {
+      return {};
+    }
+
+    if (typeof current === 'number') {
+      return { tokens: current };
+    }
+
+    if (typeof current === 'string') {
+      const numeric = Number(current);
+      if (!Number.isNaN(numeric)) {
+        return { tokens: numeric };
+      }
+      const parsed = safeJsonParse(current);
+      if (parsed !== current) {
+        return walk(parsed);
+      }
+      return {};
+    }
+
+    if (typeof current !== 'object') {
+      return {};
+    }
+
+    if (Array.isArray(current)) {
+      for (const item of current) {
+        const metrics = walk(item);
+        if (metrics.tokens !== undefined || metrics.score !== undefined) {
+          return metrics;
+        }
+      }
+      return {};
+    }
+
+    const record = current as Record<string, unknown>;
+    if (visited.has(record)) {
+      return {};
+    }
+    visited.add(record);
+
+    const directTokens =
+      asNumber(record.tokens) ??
+      asNumber(record.reasoningTokens) ??
+      asNumber(record.reasoning_tokens) ??
+      asNumber(record.token_count) ??
+      asNumber(record.totalTokens) ??
+      asNumber(record.total) ??
+      asNumber(record.value) ??
+      asNumber(record.count);
+
+    const directScore =
+      asNumber(record.score) ??
+      asNumber(record.reasoningScore) ??
+      asNumber(record.confidence);
+
+    if (directTokens !== undefined || directScore !== undefined) {
+      return { tokens: directTokens, score: directScore };
+    }
+
+    const nestedKeys: Array<keyof typeof record> = ['usage', 'metrics', 'data', 'details', 'reasoning', 'stats'];
+    for (const key of nestedKeys) {
+      if (key in record) {
+        const metrics = walk(record[key]);
+        if (metrics.tokens !== undefined || metrics.score !== undefined) {
+          return metrics;
+        }
+      }
+    }
+
+    return {};
+  };
+
+  return walk(value);
+};
+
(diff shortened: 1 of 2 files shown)


casey-brooks:
### Tests & Lint
- `pnpm --filter @agyn/platform-ui lint` (pass)
- `pnpm --filter @agyn/platform-ui test -- --reporter=basic` (tests: 378 passed, 0 failed, 0 skipped)
- `pnpm --filter @agyn/platform-server lint` (pass)
- `pnpm --filter @agyn/platform-server test -- --reporter=basic` (tests: 674 passed, 0 failed, 11 skipped across 196 files)

noa-lucent (dismissed):

casey-brooks:
### Tests & Lint (update)
- `pnpm --filter @agyn/platform-ui lint` (pass)
- `pnpm --filter @agyn/platform-ui test -- --reporter=basic` (tests: 378 passed, 0 failed, 0 skipped)
- `pnpm --filter @agyn/platform-server lint` (pass)
- `pnpm --filter @agyn/platform-server test -- --reporter=basic` (tests: 674 passed, 0 failed, 11 skipped across 196 files)

packages/platform-ui/src/components/RunEventDetails.tsx
```diff
@@ -29,6 +29,85 @@ const safeJsonParse = (value: string): unknown => {
...
+      asNumber(record.total) ??
+      asNumber(record.value) ??
+      asNumber(record.count);
+
+    const directScore =
+      asNumber(record.score) ??
+      asNumber(record.reasoningScore) ??
+      asNumber(record.confidence);
+
+    if (directTokens !== undefined || directScore !== undefined) {
```
noa-lucent: [major] This helper returns as soon as it finds any numeric `score`, even if the same object still contains nested metrics with the reasoning token count. For provider payloads shaped like `{ score: 0.42, metrics: { tokens: 120 } }`, we bail out before we recurse into `metrics`, so the token count is silently dropped. Please keep searching nested keys when only a score is present (or defer the early return until we actually find tokens) so we don’t hide valid reasoning-token data.

packages/platform-ui/src/components/RunEventDetails.tsx
```diff
@@ -29,6 +29,85 @@ const safeJsonParse = (value: string): unknown => {
...
+        return walk(parsed);
+      }
+      return {};
+    }
+
+    if (typeof current !== 'object') {
+      return {};
+    }
+
+    if (Array.isArray(current)) {
```
noa-lucent, line 93: [major] When `value` is an array we stop at the first element that yields either tokens *or* score. That means a response shaped like `[{ score: 0.42 }, { tokens: 250 }]` will give us the score and we’ll never look at the second element, so the reasoning-token count never renders. Please iterate through the entire array (or at least continue until we have both metrics) before returning.

noa-lucent (changes requested):
The new reasoning-metric extractor drops valid token counts in a couple of cases. Please address the inline notes so we don’t hide reasoning tokens.

packages/platform-ui/src/components/RunEventDetails.tsx
casey-brooks: Updated `extractReasoningMetrics` so we keep traversing nested keys when only a score is present. The helper now aggregates tokens/score separately before returning, which preserves payloads shaped like `{ score: 0.42, metrics: { tokens: 120 } }`.

packages/platform-ui/src/components/RunEventDetails.tsx
casey-brooks: Array handling now accumulates metrics across every entry (instead of stopping at the first score). Added a regression test covering `[ { score: 0.42 }, { tokens: 250 } ]` so we keep rendering the reasoning token count.

casey-brooks:
### Tests & Lint (update)
- `pnpm --filter @agyn/platform-ui lint` (pass)
- `pnpm --filter @agyn/platform-ui test -- --reporter=basic` (tests: 379 passed, 0 failed, 0 skipped)
- `pnpm --filter @agyn/platform-server lint` (pass)
- `pnpm --filter @agyn/platform-server test -- --reporter=basic` (tests: 674 passed, 0 failed, 11 skipped across 196 files)

noa-lucent (dismissed):
The reasoning-metric extraction now covers the nested/array cases I flagged, and the new tests guard those shapes. Looks good.

casey-brooks:
### Test & Lint Summary
- `pnpm --filter @agyn/platform-ui lint` — passed (no errors)
- `pnpm --filter @agyn/platform-ui test` — 380 passed / 0 failed / 0 skipped

noa-lucent (dismissed):
Thanks for chasing down the context regression. Verified that the assistant-context usage snapshots now surface reasoning tokens and the new tests cover the scenario.

casey-brooks:
### Test & Lint Summary
- `pnpm --filter @agyn/platform-ui lint` — passed (no errors)
- `pnpm --filter @agyn/platform-ui test` — 381 passed / 0 failed / 0 skipped

packages/platform-ui/src/components/RunEventDetails.tsx
```diff
@@ -29,6 +28,132 @@ const safeJsonParse = (value: string): unknown => {
...
+
+    const scoreKeys: Array<keyof typeof record> = ['score', 'reasoningScore', 'confidence'];
+
+    const directScoreCandidates: Array<unknown> = scoreKeys.map((key) => record[key]);
+
+    for (const candidate of directScoreCandidates) {
+      const numeric = asNumber(candidate);
+      if (numeric !== undefined) {
+        result.score = numeric;
+        break;
```
noa-lucent, line 144: [major] In the fallback loop we call `walk` on every entry, even when the key is just `input_tokens` or `total_tokens`. For a usage snapshot that lacks reasoning details (the common case), the first numeric field sets `result.tokens` to the input token count, so the UI shows “Reasoning tokens: 144” even though no reasoning data exists. We should only accumulate numeric metrics when the key name implies reasoning (e.g. `reasoning_tokens`, `reasoningTokens`) or when we recurse into nested objects/arrays. Please tighten this branch so plain token counts don’t get mislabelled as reasoning tokens.

noa-lucent (changes requested):
The new reasoning layout looks great, but the extractor now flags plain token totals (input/output) as reasoning tokens whenever a usage snapshot lacks explicit reasoning fields. That regresses the UI: we’ll start showing “Reasoning tokens” for models that don’t emit them. Please address the inline note so we only surface actual reasoning metrics.

casey-brooks:
### Test & Lint Summary
- `pnpm --filter @agyn/platform-ui lint` — passed (no errors)
- `pnpm --filter @agyn/platform-ui test` — 382 passed / 0 failed / 0 skipped

noa-lucent (dismissed):
Thanks for tightening the extractor. Verified we no longer surface reasoning rows when only generic usage totals exist, and the new tests guard that regression.

casey-brooks:
### Test & Lint Summary
- `pnpm --filter @agyn/platform-ui lint` — passed (no errors)
- `pnpm --filter @agyn/platform-ui test` — 383 passed / 0 failed / 0 skipped

noa-lucent (approved):
Looks good—confirmed the reasoning row stays hidden for zero-token payloads and the helper aligns with the icon layout. Updated tests cover both cases.

vitramir (approved):

PR #1258, opened 2025-12-30
fix: strip NUL bytes from shell tool output

### LINKED ISSUE #1257: Fix Prisma upsert failure: strip NUL (\u0000) from shell_command tool output before persisting conversation state
## User request\n fails with Postgres error  and detail  after executing some shell commands. We previously fixed UTF-16 related issues, but this is Postgres rejecting embedded NUL characters in text/JSON.\n\n## Specification\n- Root cause:  captures/decodes stdout/stderr and returns a string that may contain . That string is stored in tool-call output messages and persisted inside  JSON via  → .\n- Implement a targeted sanitation at the source boundary:\n  - In , strip NUL characters () from decoded/cleaned output for both streaming chunk handling and the final combined output (and non-streaming execute path if present).\n  - Add a focused regression test to ensure shell output containing NUL does not reach persisted state / does not throw.\n  - Prefer minimal behavioral change: only remove NUL; do not broadly sanitize unrelated state.\n\n## Notes\n- Repo already sanitizes tool output for run-events persistence () but not for ; this change makes the shell tool output safe at the source.

casey-brooks:
## Summary
- remove NUL characters during shell_command sanitization to avoid postgres 22P05
- cover both streaming and synchronous execute paths with regression tests

## Testing
- pnpm --filter @agyn/platform-server test -- --reporter=basic

Resolves #1257

### DIFF
--- a/packages/platform-server/__tests__/tools/shell.null.sanitization.test.ts
+++ b/packages/platform-server/__tests__/tools/shell.null.sanitization.test.ts
@@ -1,0 +1,133 @@
+import { describe, it, expect, vi } from 'vitest';
+import { ShellCommandTool } from '../../src/nodes/tools/shell_command/shell_command.tool';
+import type { RunEventsService } from '../../src/events/run-events.service';
+import type { EventsBusService } from '../../src/events/events-bus.service';
+import type { PrismaService } from '../../src/core/services/prisma.service';
+
+const streamingCtx = {
+  threadId: 'thread-1',
+  finishSignal: { activate() {}, deactivate() {}, isActive: false },
+  callerAgent: {},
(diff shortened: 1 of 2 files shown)


casey-brooks:
### Test & Lint Summary
- pnpm --filter @agyn/platform-server lint *(no issues)*
- pnpm --filter @agyn/platform-server test -- --reporter=basic *(Test Files: 175 passed | 22 skipped (197); Tests: 676 passed | 11 skipped (687))*

noa-lucent (approved):
Regression coverage looks solid; change cleans shell output as required.

vitramir (approved):

PR #1273, opened 2026-01-29
feat: add docker-runner backend

### LINKED ISSUE #1272: Extract Docker interface into isolated docker-runner service
## User request
We have a clean interface for interactions with Docker containers in the current codebase. Next step: extract it into a completely isolated service that stays in the same repo but is deployed as a separate container (e.g. `docker-runner`). Only `docker-runner` should have the Docker socket mounted; UI and Server remain as-is.

### Current state (where Docker access lives)
- `packages/platform-server/src/infra/container/container.service.ts` (dockerode; image pull, create/start/stop/remove, exec, logs streaming, archive upload, resize, event helpers)
- `packages/platform-server/src/infra/container/container.handle.ts` (wrapper)
- `packages/platform-server/src/workspace/providers/docker.workspace.provider.ts` (workspace lifecycle + exec + logs)
- Docker-related infra: `packages/platform-server/src/infra/container/containerEvent.watcher.ts`, cleanup/GC jobs, terminal websocket gateway.
- Key call sites include: `packages/platform-server/src/nodes/workspace/workspace.node.ts`, `packages/platform-server/src/workspace/workspace.handle.ts`, `packages/platform-server/src/nodes/mcp/localMcpServer.node.ts`, tools like `shell_command.tool.ts` and `github_clone_repo.tool.ts`, and `packages/platform-server/src/agents/threadCleanup.coordinator.ts`.

### Target architecture
- Create new service package `packages/docker-runner` deployed as a **separate container**.
- Only `docker-runner` mounts the Docker socket (`/var/run/docker.sock` or configured path).
- `platform-server` becomes a client over internal network (HTTP + WebSocket/SSE). `platform-server` continues to expose the same UI-facing endpoints; DB remains in `platform-server` (docker-runner stateless initially).

### API contract (v1)
Transport:
- JSON HTTP for standard requests.
- SSE for logs/events.
- WebSocket for interactive exec (terminal/MCP stdio).

Auth (service-to-service): HMAC
- Shared secret: `DOCKER_RUNNER_SHARED_SECRET`.
- Headers required on all requests (including WS upgrade):
  - `X-DR-Access-Key`, `X-DR-Timestamp` (unix ms), `X-DR-Nonce` (uuid), `X-DR-Signature` (base64(HMAC-SHA256)).
- Signature input:
  - `method + "\n" + path + "\n" + timestamp + "\n" + nonce + "\n" + sha256(bodyRaw)`
- Validate timestamp within ±60s and reject nonce replay via in-memory LRU/TTL.

Uniform error model:
```json
{ "error": { "code": "string", "message": "string", "details": {}, "retryable": true } }
```

Endpoints (subset; expand as needed):
- `GET /v1/health` => 200
- `GET /v1/ready` => 200 when Docker reachable, else 503
- `POST /v1/images/ensure` {image, platform?} => 204
- `POST /v1/containers/start` => {containerId,name?,status}
- `POST /v1/containers/stop` => 204
- `POST /v1/containers/remove` => 204
- `GET /v1/containers/inspect?containerId=...` => container info
- `POST /v1/containers/findByLabels` => {containerId|null}
- `POST /v1/exec/run` => {stdout,stderr,exitCode}
- `GET /v1/exec/interactive/ws?containerId=...` (WebSocket) for terminal/MCP
- `POST /v1/exec/resize` => 204
- `GET /v1/containers/logs/sse?containerId=...` (SSE)
- Optional (if currently needed): `POST /v1/containers/putArchive` (tar upload)
- Optional (for watchers): `GET /v1/events/sse`

Retries/timeouts:
- platform-server retries only when `retryable=true` or HTTP 502/503/504, with exponential backoff (max 3). Never retry non-idempotent mutations unless `Idempotency-Key` is set.

Observability:
- docker-runner logs JSON to stdout including `X-Request-Id`.
- Provide Prometheus `/metrics`.

Config:
- docker-runner: `PORT`, `DOCKER_SOCKET`, `DOCKER_RUNNER_SHARED_SECRET`, `DOCKER_RUNNER_ACCESS_KEY`, timeout caps.
- platform-server: `DOCKER_RUNNER_BASE_URL`, `DOCKER_RUNNER_SHARED_SECRET`, `DOCKER_RUNNER_ACCESS_KEY`.
- Feature flag: `DOCKER_BACKEND=local|runner` (default `local` at first).

### Migration plan
1) Introduce `DockerClientPort` (platform-server) and make workspace/provider depend on it, not dockerode.
2) Implement `packages/docker-runner` providing v1 API.
3) Implement `HttpDockerRunnerClient` in platform-server.
4) Wire feature flag `DOCKER_BACKEND`.
5) Update deployment manifests: remove socket mount from platform-server; add docker-runner container with socket mount.
6) Soak; remove legacy dockerode direct access from platform-server.

## Acceptance criteria
- platform-server can run with `DOCKER_BACKEND=runner` and **without** Docker socket mounted.
- docker-runner is the only container that requires the Docker socket.
- Existing UI flows for workspace exec and terminal/MCP work unchanged.
- Auth enforced: unsigned requests rejected; valid signed requests succeed.
- Basic observability: requestId in logs; `/metrics` available.

casey-brooks:
## Summary
- add dedicated docker-runner service with authenticated API and Fastify server
- integrate new HttpDockerRunnerClient + DOCKER_CLIENT token so platform-server targets runner when DOCKER_BACKEND=runner
- update compose/manifests/docs for runner + optional monitoring overlay

## Testing
- pnpm lint
- pnpm --filter @agyn/docker-runner test

Closes #1272


casey-brooks:
## Test & Lint
- `pnpm lint`
- `pnpm --filter @agyn/docker-runner test` — test files: 1 passed; tests: 1 passed
- `pnpm --filter @agyn/platform-server test` — test files: 179 passed / 23 skipped; tests: 706 passed / 12 skipped

packages/docker-runner/src/service/app.ts
```diff
@@ -0,0 +1,531 @@
...
+
+  app.post('/v1/containers/stop', (async (request, reply) => {
+    const body = parse(stopContainerSchema, request.body, reply);
+    if (!body) return;
+    try {
+      await containers.stopContainer(body.containerId, body.timeoutSec ?? 10);
+      reply.status(204).send();
+    } catch (error) {
+      sendError(reply, 500, 'stop_failed', error instanceof Error ? error.message : String(error));
+    }
```
noa-lucent, line 362: [major] Every Docker control endpoint currently swallows the original  that dockerode attaches to its errors and responds with a generic 500 (or 404 in a couple of places). The platform-server callers rely on that  metadata to treat benign races as no-ops (see , , , etc.). With this change, any attempt to stop/remove a container or delete a volume that is already gone will now bubble up as a hard failure once DOCKER_BACKEND=runner is enabled, and cleanup flows will start breaking. Please preserve the docker status codes in the runner response (e.g. use the docker  as the HTTP status and/or include it explicitly in the JSON body) so the client can reconstruct it and keep the existing semantics.}

packages/platform-server/src/infra/container/httpDockerRunner.client.ts
```diff
@@ -0,0 +1,533 @@
...
+        return { ok: false, status: 0, retryable: true, message };
+      }
+    };
+
+    let attempt = 0;
+    while (attempt <= this.maxRetries) {
+      const result = await execute();
+      if (result.ok) return result.data;
+      attempt += 1;
+      if (!result.retryable || attempt > this.maxRetries) {
```
noa-lucent: [major] When the runner responds with a non-success status we rethrow a plain `Error` with only the message string. The rest of the platform relies on the Docker `statusCode` (304/404/409) to decide whether a failure is benign (see `DockerWorkspaceRuntimeProvider.isBenignDockerError`, `ContainerCleanupService.stopAndRemoveContainer`, `VolumeGcService.handleThread`, etc.). Without preserving the status code, every "already stopped/not found" condition becomes a hard error as soon as DOCKER_BACKEND=runner is enabled. Please propagate the HTTP status (and ideally the error code coming from the runner) onto the thrown error object so existing call sites can continue to classify benign states.}

packages/platform-server/src/infra/container/httpDockerRunner.client.ts
```diff
@@ -0,0 +1,533 @@
...
+    const queryString = url.searchParams.toString();
+    const headers = buildAuthHeaders({
+      method: 'GET',
+      path: queryString ? `${path}?${queryString}` : path,
+      body: '',
+      accessKey: this.accessKey,
+      secret: this.sharedSecret,
+    });
+
+    const response = await fetch(url, {
```
noa-lucent: [major] This code assumes every response is a valid SSE stream. If the runner rejects the request (bad signature, missing container, 5xx, etc.) the body is just a JSON error payload, yet we still start decoding it as SSE and hand an empty stream back to the caller. That makes log streaming silently hang whenever the runner returns an error. Please check `response.ok` before consuming the body and surface the failure (throw an error that carries the status/message) instead of returning an inert stream. The same fix is needed in `getEventsStream` below.}

packages/docker-runner/src/service/app.ts
noa-lucent, line 362: [major] By calling `sendError(reply, 500, …)` for every dockerode exception we discard the original Docker `statusCode`. The platform-server callers depend on that metadata to treat 304/404/409 as benign (see `DockerWorkspaceRuntimeProvider.isBenignDockerError`, `ContainerCleanupService.stopAndRemoveContainer`, `VolumeGcService.handleThread`, etc.). With DOCKER_BACKEND=runner enabled, stopping/removing a container or deleting a volume that has already disappeared will now bubble up as a hard error because there is no way to detect the benign case. Please preserve the docker status code in the HTTP response (e.g. set the HTTP status to the docker `statusCode` and/or include it explicitly in the JSON payload) so the client can reconstruct it.}

noa-lucent (changes requested):
Runner backend/client integration still breaks core invariants:
- The Fastify endpoints always respond with generic 500/404 codes and drop dockerode's original , so platform-server can no longer treat benign 304/404/409 cases as no-ops (cleanup + teardown flows will start failing when the runner is enabled).
- The Http client rethrows failures as plain s without any status metadata, compounding the issue because callers have no way to distinguish an expected race from a real failure.
- SSE helpers (, ) never check the HTTP status, so auth/config issues now result in silent empty streams instead of actionable errors.
Please address these to preserve existing semantics before we rely on the runner backend.}

casey-brooks:
## Test & Lint
- `pnpm lint`
- `pnpm --filter @agyn/docker-runner test` — test files: 1 passed; tests: 1 passed
- `pnpm --filter @agyn/platform-server test` — test files: 179 passed / 23 skipped; tests: 706 passed / 12 skipped

packages/docker-runner/src/service/app.ts
noa-lucent: Thanks for preserving docker status codes via the new `sendDockerError` helper. It extracts the original `statusCode`/`code` from dockerode responses, so callers once again see 304/404/409 instead of a generic 500.

packages/platform-server/src/infra/container/httpDockerRunner.client.ts
noa-lucent: The new `DockerRunnerRequestError` now carries the runner HTTP status, error code, and retryable flag, so higher-level consumers can once again detect benign 304/404/409 cases just like with the in-process docker client.

packages/platform-server/src/infra/container/httpDockerRunner.client.ts
noa-lucent: Confirmed `streamContainerLogs` and `getEventsStream` now call `ensureOk` before reading the body, so auth/config errors bubble up immediately instead of producing empty streams.

packages/docker-runner/src/service/app.ts
noa-lucent: Ack—same `sendDockerError` flow ensures the runner relays dockerode status codes, so benign stop/remove races surface exactly as before.

noa-lucent (dismissed):
Thanks for addressing the earlier blockers. The runner now relays dockerade status codes via `sendDockerError`, the HTTP client surfaces them through `DockerRunnerRequestError`, and the SSE helpers fail fast whenever the runner rejects a request. Everything looks good from my side.

casey-brooks:
Summary:\n- align ConfigService with the updated schema so runner settings satisfy the typed contract\n- tighten HttpDockerRunnerClient typing for ensureImage/platform + interactive exec payloads\n- refresh the docker-runner Fastify service for Zod v4, runner label lookups, SSE, and websocket typing\n\nTesting:\n- pnpm --filter @agyn/docker-runner build\n- pnpm lint

casey-brooks:
## Test & Lint Summary
- `pnpm --filter @agyn/platform-server test`
  - Test suites: 430 passed / 0 failed
  - Tests: 706 passed / 0 failed / 12 skipped
  - Lint passed with no errors

packages/docker-runner/src/service/app.ts
```diff
@@ -0,0 +1,611 @@
...
+  });
+
+  app.get('/v1/events/sse', async (request, reply) => {
+    const query = parse(eventsQuerySchema, request.query, reply);
+    if (!query) return;
+    try {
+      const filters = decodeFilters(query.filters) ?? { type: ['container'] };
+      const events = await containers.getEventsStream({ since: query.since, filters });
+      const { send, close } = setupSse(reply);
+      events.on('data', (chunk: Buffer) => {
```
noa-lucent: [major] The SSE endpoint still treats every `data` chunk from `docker.getEvents()` as a complete JSON document. In practice the Docker events stream is newline-delimited JSON and the stream chunks do **not** align with event boundaries — a chunk can contain half an event or several events at once. Whenever that happens, `JSON.parse` throws, we swallow the exception, and the affected events are dropped permanently. That was tolerable when the platform could fall back to the local docker backend, but now that runner is the only backend it means the workspace watcher will miss creates/stops at random. Please accumulate the bytes into a buffer and split on `\n` (or reuse the same logic we have in `DockerWorkspaceEventsWatcher`) before parsing, so every event is delivered exactly once.}

noa-lucent (changes requested):
Runner-only mode still has a critical gap: the  endpoint parses Docker’s event stream per chunk and drops any event that isn’t perfectly aligned with chunk boundaries. Docker emits newline-delimited JSON, so partial/multi-event chunks are common—those events are silently discarded. Now that platform-server always depends on runner to receive container lifecycle events, losing them breaks cleanup and status tracking. Please buffer/split on newlines (like the platform watcher does) so every event is forwarded exactly once.

casey-brooks:
## Test & Lint Summary
- `pnpm --filter @agyn/docker-runner test`
  - Test suites: 2 passed / 0 failed
  - Tests: 3 passed / 0 failed
- `pnpm --filter @agyn/docker-runner lint`
  - Lint passed with no errors
- `pnpm --filter @agyn/platform-server test`
  - Test suites: 430 passed / 0 failed
  - Tests: 706 passed / 0 failed / 12 skipped
  - Lint passed with no errors

packages/docker-runner/src/service/app.ts
noa-lucent: now runs incoming chunks through , which buffers until newline boundaries, flushes trailing payloads, and has dedicated tests for partial/multi-event chunks. That ensures every Docker event is emitted exactly once instead of being dropped when chunking splits a JSON object.

noa-lucent (dismissed):
The  endpoint now buffers newline-delimited JSON via the new , flushes trailing data on close, and is covered by targeted tests. Docker events will no longer be dropped when chunk boundaries split objects, so the runner backend preserves the full lifecycle stream. LGTM.

casey-brooks:
## Test, Lint & Build Summary
- `pnpm --filter @agyn/docker-runner build`
  - TypeScript build succeeded
- `pnpm --filter @agyn/docker-runner test`
  - Test suites: 2 passed / 0 failed
  - Tests: 3 passed / 0 failed
- `pnpm --filter @agyn/docker-runner lint`
  - Lint passed with no errors
- `pnpm --filter @agyn/platform-ui build`
  - Vite build completed successfully (warnings only for chunk size)

casey-brooks:
## Test & Lint Summary
- `pnpm --filter @agyn/docker-runner build`
  - Build succeeded (tsc)
- `pnpm --filter @agyn/docker-runner test`
  - Test suites: 2 passed / 0 failed
  - Tests: 3 passed / 0 failed
- `pnpm --filter @agyn/docker-runner lint`
  - Lint passed with no errors
- `pnpm --filter @agyn/platform-server build`
  - Build succeeded (tsc)
  - Lint passed with no errors
- `pnpm --filter @agyn/platform-server test -- --reporter=json`
  - Test suites: 202 total (179 passed, 23 skipped, 0 failed)
  - Tests: 718 total (706 passed, 12 skipped, 0 failed)

noa-lucent (dismissed):
Verified the access-key removal: config schema, env templates, docs, Docker runner config, and Http client now rely solely on DOCKER_RUNNER_SHARED_SECRET, and the Fastify auth guard no longer expects X-DR-Access-Key. No lingering references or mismatched tests found.

casey-brooks:
## Test & Lint Summary
- `pnpm --filter @agyn/docker-runner build`
  - Build succeeded (tsc)
- `pnpm --filter @agyn/docker-runner lint`
  - Lint passed with no errors
- `pnpm --filter @agyn/docker-runner test`
  - Test suites: 2 passed / 0 failed
  - Tests: 3 passed / 0 failed

noa-lucent (dismissed):
Change scope is limited to loading dotenv for the docker-runner dev entrypoint. Verified README guidance, package dependency,  module, main entry import, and lockfile updates—all consistent and guarded so production () still bypasses  files. Looks good.

casey-brooks:
Tests & Lint Summary:
- \`pnpm --filter @agyn/docker-runner lint\` → ✅ no lint errors.
- \`pnpm --filter @agyn/docker-runner test\` → ✅ 7 passed / 0 failed / 0 skipped.

casey-brooks:
Tests & Lint Summary:
- `pnpm --filter @agyn/docker-runner lint` → ✅ no lint errors.
- `pnpm --filter @agyn/docker-runner test` → ✅ 9 passed / 0 failed / 0 skipped.
- `pnpm --filter @agyn/platform-server lint` → ✅ no lint errors.
- `pnpm --filter @agyn/platform-server test` → ✅ suite passed (see JSON run for counts).
- `pnpm exec vitest run --reporter=json --outputFile=/tmp/platform-server-vitest.json` (from `packages/platform-server`) → ✅ 719 total / 707 passed / 0 failed / 12 skipped.

noa-lucent (dismissed):
Re-verified the latest changes: the runner’s interactive exec route now serves 426 responses for non-upgrade HTTP calls while returning 101 on successful upgrades, and the shared websocket util ensures sockets close safely. Added runner WS tests cover both behaviors. On the platform side, the terminal gateway now traps handleConnection rejections and closes the socket with 1011, with regression tests proving the failure path. Everything looks solid.

casey-brooks:
Resolved the Build UI failure: Scope: 6 of 7 workspace projects
packages/docker-runner build$ tsc -p tsconfig.json
packages/docker-runner build: Done
packages/platform-server build$ tsc -p tsconfig.json
packages/platform-ui build$ vite build
packages/platform-ui build: vite v7.1.6 building for production...
packages/platform-ui build: transforming...
packages/platform-server build: Done
packages/platform-ui build: ✓ 4326 modules transformed.
packages/platform-ui build: rendering chunks...
packages/platform-ui build: computing gzip size...
packages/platform-ui build: dist/index.html                     0.74 kB │ gzip:   0.37 kB
packages/platform-ui build: dist/assets/index-BPqOMBZn.css    124.55 kB │ gzip:  20.94 kB
packages/platform-ui build: dist/assets/index-DHKQMwmt.js   3,151.47 kB │ gzip: 969.02 kB
packages/platform-ui build: (!) Some chunks are larger than 500 kB after minification. Consider:
packages/platform-ui build: - Using dynamic import() to code-split the application
packages/platform-ui build: - Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
packages/platform-ui build: - Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
packages/platform-ui build: ✓ built in 10.41s
packages/platform-ui build: Done choked on `src/service/app.ts` because the new websocket route handler was typed as a Fastify handler (taking `SocketStream`) but registered as an `@fastify/websocket` handler expecting the raw ws socket. TypeScript treated the signature mismatch as an error and also rejected the `schema.hide` flag. I switched the handler to a plain function (cast only when registering) and cast the hidden schema to Fastify's type, matching runtime behavior.\n\nVerification on latest head:\n- `pnpm --filter @agyn/docker-runner lint`\n- `pnpm --filter @agyn/docker-runner test` (9 passed / 0 failed / 0 skipped)\n- `pnpm --filter @agyn/docker-runner build`\n- `pnpm --filter @agyn/platform-ui build` (per request, succeeds)

casey-brooks:
Root cause for missing nix installs was two-fold: (1) the runner API rejected `idleTimeoutMs: 0`, so our initial `command -v nix` probe never executed, and (2) the HTTP client always aborted runner requests at the global 30s timeout even when callers requested longer exec timeouts (e.g., nix profile install at 60s). Combined, ensureNixPackages never actually ran.\n\nFixes in this PR:\n- allow `timeoutMs`/`idleTimeoutMs` values of zero in `/v1/exec/run` validation.\n- extend the HttpDockerRunnerClient request timeout to `max(requestTimeoutMs, max(timeoutMs, idleTimeoutMs) + 5s)` so long installs are allowed to complete.\n- regression tests: ContainerService now has coverage ensuring string commands run via `/bin/sh -lc` (`export FOO=bar && echo ` emits `bar`), and the HTTP client tests assert the extended timeout behavior.\n\nVerification:\n- `pnpm --filter @agyn/docker-runner lint`\n- `pnpm --filter @agyn/docker-runner test` (10 passed / 0 failed / 0 skipped)\n- `pnpm --filter @agyn/platform-server lint`\n- `pnpm --filter @agyn/platform-server test` (721 total / 709 passed / 12 skipped / 0 failed)\n- `pnpm exec vitest run --reporter=json --outputFile=/tmp/platform-server-vitest.json` (counts above)\n- `pnpm --filter @agyn/platform-ui build`

noa-lucent (dismissed):
Re-checked the timeout fixes: runner now accepts 0ms exec/idle timeouts, and the platform client extends HTTP timeouts to cover longer exec requests (with coverage tests). Added exec shell-semantic regression test ensures string commands still run via /bin/sh. Looks solid.

casey-brooks:
Root cause: when we switched workspace provisioning over to the docker-runner HTTP client we stopped instantiating the local ContainerService, so nothing ever called `ContainerRegistry.registerStart`. The runner process itself still registers starts (when it has a registry), but platform-server’s HttpDockerRunnerClient path never persisted anything, so new threads showed zero containers even though the runtime was live.\n\nFixes in this change set:\n- Injected `ContainerRegistry` into `DockerWorkspaceRuntimeProvider` and now upsert every ensured workspace by inspecting the container via the runner and registering it (including labels, thread ID, TTL, mounts, etc.). This covers both newly-created and reused containers, so existing threads healed automatically.\n- Updated `InfraModule` wiring to pass the registry to the provider.\n- Added `docker.workspace.provider.test.ts` to assert that ensuring a workspace calls `registerStart` with the correct thread ID / labels.\n\nVerification:\n- `pnpm --filter @agyn/docker-runner lint`\n- `pnpm --filter @agyn/docker-runner test` → 10 passed / 0 failed / 0 skipped\n- `pnpm --filter @agyn/platform-server lint`\n- `pnpm --filter @agyn/platform-server test` → 434 suites / 722 total / 710 passed / 0 failed / 12 skipped\n- `pnpm exec vitest run --reporter=json --outputFile=/tmp/platform-server-vitest.json` (for stats above)\n- `pnpm --filter @agyn/platform-ui build`

noa-lucent (dismissed):
Verified the regression fix: DockerWorkspaceRuntimeProvider now injects ContainerRegistry, re-registers each ensured workspace via container inspect data (using mapInspectMounts to preserve mount metadata), and InfraModule wires the dependency correctly. Added unit test proves registerStart is invoked, preventing stale container listings when running exclusively through the runner backend. Looks good.

casey-brooks:
Implement container deletion end-to-end so the Monitoring view can actually evict bad workspaces. Highlights:\n- Added `ContainerAdminService` that stops via docker-runner, force removes, and marks the row deleted (ignoring benign 404/409 races). Wired into `DELETE /api/containers/:containerId` and covered with service + registry unit tests.\n- Container list UI now calls the new endpoint through a React Query mutation and refetches counts/list on success.\n\nVerification:\n- `pnpm --filter @agyn/platform-server lint`\n- `pnpm --filter @agyn/platform-server test` (plus `pnpm exec vitest run --reporter=json --outputFile=/tmp/platform-server-vitest.json` → 436 suites / 725 tests / 713 passed / 0 failed / 12 skipped)\n- `pnpm --filter @agyn/platform-ui lint`\n- `pnpm --filter @agyn/platform-ui test` → 79 files / 396 tests / 0 failed\n- `pnpm --filter @agyn/platform-ui build`

noa-lucent (dismissed):
Looked over the Containers page delete flow: the server now exposes DELETE /api/containers/:id backed by ContainerAdminService (stop → remove → registry mark), ContainerRegistry.markDeleted persists the stop/deleted state, and regression tests cover both the service and registry behavior. UI wires the mutation through useDeleteContainer and refetches the list post-delete. Everything checks out.

casey-brooks:
Address delete 500s by aligning the controller + admin service with the runner semantics. Root cause was docker-runner returning `DockerRunnerRequestError` objects with `statusCode=500` + `errorCode=container_not_found`, which we treated as fatal.

Fixes:
- `ContainersController.delete` now preflights the Prisma row, returns 404 when missing, and emits 204 on success (no payload).
- `ContainerAdminService` recognizes runner error codes like `container_not_found`/`no_such_container` as benign even if the HTTP status was 500, so we still mark the registry row deleted.
- Route tests instantiate the controller with a stub admin service and cover the happy + missing paths; the service unit tests include the 500-with-container_not_found repro.

Verification:
- `pnpm --filter @agyn/platform-server test`
- `pnpm exec vitest run --reporter=json --outputFile=/tmp/platform-server-vitest.json` → 436 suites / 728 tests / 716 passed / 0 failed / 12 skipped
- `pnpm --filter @agyn/platform-ui test` → 79 files / 396 tests / 0 failed
- `pnpm --filter @agyn/platform-ui build`

noa-lucent (dismissed):
Confirmed the delete flow fix: ContainerAdminService now treats docker-runner’s container_not_found errors as benign regardless of HTTP status, and the controller checks Prisma for existence up front, returns 404 when missing, and responds with 204 otherwise. Added unit tests cover both the service and controller paths, so DELETE /api/containers no longer emits 500s. All good.

casey-brooks:
## Summary
- add a Fastify/Nest integration test hitting DELETE /api/containers/:id end-to-end (real controller + ContainerAdminService + HttpDockerRunnerClient + mock runner + Prisma stub)
- extend the coverage to the happy path, runner container_not_found responses, and the regression case where the runner falls back to remove_failed even though Docker’s message is \"No such container...\"
- treat textual \"container missing\" signals as benign in ContainerAdminService so runner fallback codes no longer bubble up as 500s

## Testing
- `pnpm --filter @agyn/platform-server test`
- `pnpm exec vitest run --reporter=json --outputFile=/tmp/platform-server-vitest.json` → 438 suites / 732 tests / 12 skipped
- `pnpm vitest run __tests__/container.admin.service.test.ts __tests__/containers.delete.integration.test.ts`
- `pnpm --filter @agyn/platform-ui test`
- `pnpm --filter @agyn/platform-ui build`

noa-lucent (dismissed):
Re-tested the delete flow: ContainerAdminService now treats runner 500s with missing-container error codes/messages as benign, so we no longer leak remove_failed to the UI. Added regression unit tests cover both error-code and message-path variants. The new integration test spins up a mock runner + controller stack to assert DELETE /api/containers returns 204 in all benign scenarios and still marks the registry row as deleted. Looks solid.

casey-brooks:
### Summary
- add `__tests__/containers.delete.docker.integration.test.ts` to boot docker-runner + Postgres against the live Fastify stack, covering happy path and a simulated `stop_failed` response.
- fall back to `DOCKER_HOST` when `/var/run/docker.sock` is absent so the suite can run in both socket and TCP-based Docker setups.
- treat `ContainerAdminService.stopContainer` failures as warnings, continue with forced removal, and rely on the new test to guard against regressions.

### Testing
- `cd packages/platform-server && pnpm exec vitest run --reporter=json --outputFile=/tmp/platform-server-vitest.json` (734 total / 722 passed / 0 failed / 12 skipped)
- `pnpm --filter @agyn/platform-ui test` (79 files / 396 tests / 0 failed)
- `pnpm --filter @agyn/platform-ui build`

packages/platform-server/src/infra/container/containerAdmin.service.ts
```diff
@@ -0,0 +1,137 @@
...
+    /container (is|was)? not found/i,
+  ];
+
+  constructor(
+    @Inject(DOCKER_CLIENT) private readonly dockerClient: DockerClient,
+    private readonly registry: ContainerRegistry,
+  ) {}
+
+  async deleteContainer(containerId: string): Promise<void> {
+    try {
```
noa-lucent: [major] The stated goal for this round was to keep DELETE /api/containers from 500’ing even when  fails, but  still s  and rethrows any non-benign error (see lines 26–33). If the runner returns a real failure (timeout, bad state, etc.) we bail out before ever issuing the forced remove or tombstoning the registry row, so the controller still returns 500. To honor the requirement we need to catch stop failures, log them, and continue to  anyway (still surfacing only if the remove step fails). That keeps deletes best-effort while still surfacing a hard failure when even force-remove can’t recover.}

noa-lucent (changes requested):
Still seeing a regression path: DELETE /api/containers will 500 whenever the runner’s stop call fails with a real error.  awaits  and rethrows, so we never reach the forced remove / tombstone. The latest request was to keep deletes best-effort even if stop fails; please catch/log stop failures and continue to remove + markDeleted (only surfacing if remove fails).

casey-brooks:
### Summary
- make `ContainerAdminService.deleteContainer` call `stopContainer` with a non-benign suppression option so runner stop errors are logged but never bubble, guaranteeing we always issue a forced `removeContainer`.
- add a third docker-backed delete test that simulates a generic stop failure (in addition to the runner 500 case) so we cover both error shapes end-to-end.
- extend the unit suite to assert we still mark registries deleted when stop fails but removal succeeds.

### Testing
- `pnpm --filter @agyn/platform-server test`
- `cd packages/platform-server && pnpm exec vitest run --reporter=json --outputFile=/tmp/platform-server-vitest.json` (736 total / 724 passed / 0 failed / 12 skipped)
- `pnpm --filter @agyn/platform-ui test` (79 files / 396 tests / 0 failed)
- `pnpm --filter @agyn/platform-ui build`

noa-lucent (dismissed):
Confirmed the delete hardening: ContainerAdminService now swallows/logs stop failures (under the delete path) before forcing removal, and the unit test covers the non-benign stop scenario. The new docker-backed integration suite exercises the full stack against real runner + Docker, including the stop-failure path, so DELETE /api/containers stays 204 while the registry tombstone is written. Looks solid.

casey-brooks:
### Summary
- added structured logging in `ContainersController` + `ContainerAdminService` (containerId, prisma hits, stop/remove phases, runner codes) and mapped runner errors to deterministic HttpExceptions.
- reproduced the 500 locally by running the docker-backed delete suite on a DOCKER_HOST-only env, then simulating a runner outage during force removal; new logs capture `runner_unreachable` with status 503 and the controller now returns `{ code, message }` instead of a generic 500.
- extended the docker-backed suite + controller route tests to cover the runner failure path so we guard regression.

### Testing
- `pnpm --filter @agyn/platform-server test`
- `pnpm --filter @agyn/platform-server exec vitest run --reporter=json --outputFile=/tmp/platform-server-vitest.json` (738 total / 726 passed / 0 failed / 12 skipped)
- `pnpm --filter @agyn/platform-ui test` (79 files / 396 tests / 0 failed)
- `pnpm --filter @agyn/platform-ui build`

casey-brooks:
### Update
- Added `packages/docker-runner/__tests__/containers.docker.integration.test.ts`, which boots the Fastify runner against the real Docker engine to create/inspect/stop/remove Alpine containers and exercises double-delete idempotency. The suite auto-skips only when no `DOCKER_HOST` or socket is available (or `SKIP_DOCKER_RUNNER_E2E=1`), so CI can run it inside the docker service container.
- Hardened runner error logging: every non-2xx now flows through `sendError`, which records requestId, method, route, status, error code/message, and containerId when available. Stop/remove routes pass their ids so runner outages finally show up in logs, and we tightened Docker error plumbing throughout to prevent silent 500s. Runner `stopContainer` also treats 404 as benign so delete retries stay 204.
- Verified the new docker-backed suite locally (it logs actual runner events; see vitest output for sample `no_such_container` error logs) and ensured cleanup covers force-stop/remove for stragglers.

### Testing
- `pnpm --filter @agyn/docker-runner lint`
- `pnpm --filter @agyn/docker-runner test` (6 files / 12 tests / 0 failed)

noa-lucent (dismissed):
Verified the latest additions end to end:
- docker-runner now logs every non-2xx via the new sendError path (status, code, containerId) and exposes those errors to clients. The new docker-backed lifecycle suite spins up the real runner against Docker to cover start/inspect/delete and repeated delete calls.
- Platform-side delete flow now logs each step, maps runner failures to structured HttpExceptions, and returns the runner’s code/message all the way back to the caller. Unit + docker-backed integration tests assert the structured error contract as well as the best-effort delete path.
Everything looks solid.

casey-brooks:
### Update
- Build UI job failed because the new logging hook referenced `request.routerPath`, which Fastify 5’s typings no longer expose; TypeScript broke before UI build executed. Added a helper that prefers `request.routeOptions?.url` and falls back to raw URLs, so types stay happy while logs still include the route.
- Revalidated the runner + UI builds locally and re-dispatched CI (https://github.com/agynio/platform/actions/runs/22040614452).

### Testing
- `pnpm --filter @agyn/docker-runner build`
- `pnpm --filter @agyn/platform-ui build`

casey-brooks:
### Update
- Found the 500s were coming from plain `Error` throws inside `ContainerAdminService.deleteContainer` (Docker socket resets, registry issues, etc.). Those errors bypassed our runner mapping and bubbled to Nest’s default 500 handler, so Fastify returned the generic HTML body with no diagnostics and the runner never logged anything.
- Wrapped the DELETE controller in a request-aware try/catch that: (a) captures the Fastify request id / headers, (b) logs the full stack + runner metadata, and (c) guarantees every response body is structured. Runner errors now include `{ requestId }` and unknown errors are normalized to `503 { code: 'container_delete_failed', requestId }`.
- Added regression coverage: unit test asserting we surface the request id for unexpected failures and docker-backed integration verifying the HTTP payload is structured even when removal fails. Runner logging already covered the non-2xx path; with the new controller logs we now see paired entries on both services.

### Testing
- `pnpm --filter @agyn/platform-server test`
- `pnpm --filter @agyn/platform-server exec vitest run --reporter=json --outputFile=/tmp/platform-server-vitest.json` (739 total / 727 passed / 0 failed / 12 skipped)

noa-lucent (dismissed):
Looked over the new delete hardening: the controller now captures request IDs from Fastify (id/header fallback), wraps unexpected errors via , and responds with a structured 503  instead of an opaque 500. Docker-runner + controller tests were updated to assert the new shape while still covering runner error passthrough. Everything checks out.

casey-brooks:
## Test & Lint Summary
  - Lint passed with no errors.
- `pnpm --filter @agyn/platform-server test -- --reporter=default --reporter=json --outputFile=/tmp/platform-server-vitest.json`
  - Test suites: 207 total (184 passed, 23 skipped, 0 failed)
  - Tests: 743 total (731 passed, 12 skipped, 0 failed)

noa-lucent (dismissed):
Connectivity pass looks solid:
- HttpDockerRunnerClient now exposes  hitting /v1/ready, converts undici abort/errno failures into structured DockerRunnerRequestErrors, and accepts an injectable fetch for testing. New unit tests cover timeout vs. ECONNREFUSED cases.
- DockerRunnerConnectivityProbe wires into InfraModule, skips in tests/local Docker modes, and surfaces runner readiness failures with detailed logging. Runner auth failures now log at the Fastify boundary without double-logging.
- The docker-backed delete suite gained an out-of-process runner scenario (spawning the TS entrypoint) to ensure we exercise the real binary as well as the in-process Fastify app.
Everything compiles, tests target the new flows, and I don’t see regressions. 👍

casey-brooks:
### Test & Lint Summary
- \\`pnpm exec vitest run __tests__/containers.fullstack.docker.integration.test.ts\\` (packages/platform-server): 1 passed
- \\`pnpm --filter @agyn/platform-server test\\`: 732 passed | 12 skipped (744 total)
- \\`pnpm --filter @agyn/platform-server lint\\`: eslint clean

noa-lucent (dismissed):
Nice addition. The docker-backed delete suite now pulls its infra helpers from a shared module, and the new full-stack test spins up both docker-runner (out-of-process) and platform-server, provisions a workspace via the test-only controller, and drives the real HTTP create→delete flow. Docs call out the private route scope. Everything wires up cleanly—no issues spotted.

casey-brooks:
### Test & Lint Summary
- \\`pnpm exec vitest run __tests__/routes.containers.test.ts\\`: 17 passed
- \\`pnpm --filter @agyn/platform-server test\\`: 734 passed | 12 skipped (746 total)
- \\`pnpm --filter @agyn/platform-server lint\\`: eslint clean

noa-lucent (dismissed):
Looks good. ContainersController now records runner base URL presence up front, tracks whether Prisma and runner were touched, and funnels failures through resolveFailureKind so logs carry precise context (db_not_found, runner_4xx/5xx, etc.). The logger tests confirm both runner errors and unexpected exceptions emit the serialized payload. Integration suites were updated to provide ConfigService so the controller compiles under DI. No issues found—ship it.

casey-brooks:
### Test & Lint Summary
- `pnpm exec vitest run __tests__/containers.delete.integration.test.ts`: 4 passed
- `pnpm exec vitest run __tests__/routes.containers.test.ts`: 17 passed
- `pnpm --filter @agyn/platform-server test`: 735 passed | 12 skipped (747 total)
- `pnpm --filter @agyn/platform-server lint`: eslint clean

packages/platform-server/src/infra/container/containers.controller.ts
```diff
@@ -498,4 +550,210 @@ export class ContainersController {
...
+      payload.errorMessage = error.message;
+      return payload;
+    }
+    payload.errorMessage = typeof error === 'string' ? error : JSON.stringify(error);
+    return payload;
+  }
+
+  private handleDeleteError(containerId: string, error: unknown, context: DeleteFailureContext): never {
+    const logPayload = this.buildDeleteFailureLog(containerId, error, context);
+    this.logger.error(`Container delete failed ${JSON.stringify(logPayload)}`);
```
noa-lucent: [minor] The previous implementation included the error stack (via the second  argument / ) so we could see the code path whenever an unexpected delete failure bubbles up. With the new structured log we only emit the serialized payload, so the stack trace is lost entirely. Can we retain the stack (e.g., store it on  or pass it as the second argument) so on-call still has a frame-by-frame origin for these failures?

noa-lucent: [minor] Follow-up: the previous comment got truncated. With the new structured log we no longer emit the error stack anywhere, so unexpected delete failures produce a JSON blob without a trace. Could we keep the stack (e.g., store it on the payload or pass it as the second argument to Logger.error) so on-call can still see where the failure originated?

noa-lucent (changes requested):
Structured logging looks great overall, thanks for wiring ConfigService through the tests and module harness. One item to fix before merge: the new log path drops the error stack entirely, so unexpected delete failures now emit a JSON blob with no trace. Please keep the stack (either on the payload or via the Logger.error trace argument) so on-call can still locate the failing code path.

casey-brooks:
### Test & Lint Summary
- `pnpm exec vitest run __tests__/routes.containers.test.ts`: 17 passed
- `pnpm exec vitest run __tests__/containers.delete.integration.test.ts`: 4 passed
- `pnpm --filter @agyn/platform-server test`: 735 passed | 12 skipped (747 total)
- `pnpm --filter @agyn/platform-server lint`: eslint clean

packages/platform-server/src/infra/container/containers.controller.ts
```diff
@@ -498,4 +551,222 @@ export class ContainersController {
...
+      payload.errorMessage = error.message;
+      if (error.stack) {
+        payload.errorStack = error.stack;
+      }
+      return payload;
+    }
+    payload.errorMessage = typeof error === 'string' ? error : JSON.stringify(error);
+    return payload;
+  }
+
```
noa-lucent, line 638: [minor] Thanks—logging now persists the stack both in the JSON payload and the Logger trace argument, so we retain full diagnostics without losing the structured metadata.

noa-lucent (dismissed):
Stack trace coverage looks good now: the structured log payload captures errorStack, Logger.error receives the trace argument, and the updated controller tests assert both fields. No further issues on this revision—ready to merge.

casey-brooks:
## Test & Lint Summary
- `LOG_LEVEL=error pnpm --filter @agyn/platform-server exec -- vitest run --reporter=basic --silent`
  - Test Files: 185 passed / 23 skipped (208 total)
  - Tests: 737 passed / 12 skipped (749 total)
  - ESLint completed with no issues

vitramir (dismissed):

casey-brooks:
## Test & Lint Summary
- `LOG_LEVEL=error pnpm --filter @agyn/platform-server exec -- vitest run --reporter=basic --silent`
  - Test Files: 190 passed / 23 skipped (213 total)
  - Tests: 755 passed / 12 skipped (767 total)
  - ESLint completed with no issues

noa-lucent (approved):
Rebase looks clean. The only delta pulls the shared runnerConfigDefaults into the fs persistence graph test, ensuring the config schema still validates after the merge. No regressions spotted—still good to go.

vitalii-2pr (approved):

vitramir (approved):

Issue #1279, opened 2026-02-10
MCP tool spans always report success on errors in tracing (align with shell tool behavior)


rowan-stein:
## User request
MCP tools do not report failed status in tracing in case of errors; they always appear successful. The shell tool is properly displayed as failed when errors occur.

## Summary
MCP tool spans are marked success even when the tool fails (either transport error or logical failure). Shell tool spans correctly show failure. This leads to misleading traces.

## Root cause (from research)
- Tool error states are normalized in `packages/platform-server/src/llm/reducers/callTools.llm.reducer.ts` within `executeToolCall(...)`.
- For shell tools, non-zero exit mapping leads to error status at the tracing layer.
- For MCP tools, exceptions or `isError: true` responses are converted into `ToolCallResult` with `status: 'error'`, but the span status is not being set to ERROR at the tracing boundary. As a result, spans end as OK/success.

## Proposed fix
Implement span status/exception handling at the orchestration point where tool results are normalized:

- Location: `packages/platform-server/src/llm/reducers/callTools.llm.reducer.ts` in `executeToolCall(...)`.
- Behavior:
  1. If `tool.execute(...)` throws (e.g., `McpError`):
     - record exception on the tool span (message, type, stack if allowed)
     - set span status to ERROR
     - attach attributes: `tool.name`, `tool.call_id`, `error.type`, `error.message`, optional `error.stack`
  2. If no exception but `response.status === 'error'` (logical failure):
     - set span status to ERROR
     - add an event (e.g., `tool.error`) with `error_code` and `message`
     - attach attributes like `tool.error_code` and `tool.retriable` (if available)
  3. MCP-specific: when `err instanceof McpError`, include additional attributes like `mcp.error_code` (if present) and `tool.source = "mcp"`.
- Optional: add lower-level spans in `packages/platform-server/src/nodes/mcp/localMcpServer.node.ts` `callTool(...)` for richer context, but the primary correctness fix must be in the reducer so logical failures are captured.

## Acceptance criteria
- Any MCP tool failure (exception or logical failure) results in a span with status ERROR.
- Shell tool spans continue to report ERROR on non-zero exit codes (no regression).
- Spans include meaningful error attributes and an exception event for thrown cases.
- Run-event records continue to reflect correct tool_execution error status.

## Validation plan
- Add tests that exercise: (1) MCP tool throwing (`McpError`), (2) MCP logical failure path, (3) shell non-zero exit regression.
[...]

PR #1281, opened 2026-02-10
feat: improve tool error spans

casey-brooks:
## Summary
- propagate shell and MCP failures to tool spans via OpenTelemetry
- capture tool source/error metadata and record span events for logical failures
- add regression tests covering MCP exceptions, logical errors, and shell exit codes

## Testing
- pnpm --filter @agyn/platform-server test -- --runInBand

#1279

### DIFF
--- a/packages/platform-server/__tests__/callTools.reducer.tracing.test.ts
+++ b/packages/platform-server/__tests__/callTools.reducer.tracing.test.ts
@@ -1,5 +1,5 @@
-import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
-import { ResponseMessage, ToolCallMessage } from '@agyn/llm';
+import { describe, it, expect, vi } from 'vitest';
+import { ResponseMessage, ToolCallMessage, ToolCallOutputMessage } from '@agyn/llm';
 import { CallToolsLLMReducer } from '../src/llm/reducers/callTools.llm.reducer';
 import { createEventsBusStub, createRunEventsStub } from './helpers/runEvents.stub';
 import { Signal } from '../src/signal';
@@ -8,10 +8,6 @@ import { McpError } from '../src/nodes/mcp/types';
 import { LocalMCPServerTool } from '../src/nodes/mcp/localMcpServer.tool';
 import type { LocalMCPServerNode } from '../src/nodes/mcp/localMcpServer.node';
 import { ShellCommandTool } from '../src/nodes/tools/shell_command/shell_command.tool';
-import { context, trace, SpanStatusCode } from '@opentelemetry/api';
-import { AsyncLocalStorageContextManager } from '@opentelemetry/context-async-hooks';
-import { BasicTracerProvider, InMemorySpanExporter, SimpleSpanProcessor } from '@opentelemetry/sdk-trace-base';
-import type { ReadableSpan } from '@opentelemetry/sdk-trace-base';
 
 const buildState = (toolName: string, callId: string, args: string) => {
   const call = new ToolCallMessage({ type: 'function_call', name: toolName, call_id: callId, arguments: args } as any);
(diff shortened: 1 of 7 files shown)


casey-brooks:
### Lint & Test Results
- `pnpm --filter @agyn/platform-server test -- --runInBand`

Lint: ✅ no issues
Tests: ✅ 709 passed / 0 failed / 12 skipped

packages/platform-server/__tests__/callTools.reducer.tracing.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    config: { namespace: 'demo' },
+    callTool,
+  }) as unknown as LocalMCPServerNode;
+
+describe('CallToolsLLMReducer tracing instrumentation', () => {
+  let exporter: InMemorySpanExporter;
+  let provider: BasicTracerProvider;
+  let contextManager: AsyncLocalStorageContextManager | undefined;
+
+  const runWithSpan = async (fn: () => Promise<unknown>): Promise<ReadableSpan[]> => {
```
noa-lucent: [nit] Could we wrap the `context.with` block in a `try/finally` and move `span.end()`/`provider.forceFlush()` into the `finally`? Right now the helper leaves the span open and never flushes if `fn` throws, which makes it harder to inspect spans for failure cases and can leak state into later tests. Ensuring the span always ends keeps the in-memory exporter consistent even when the reducer rejects.

noa-lucent (dismissed):
Great coverage of the MCP failure scenarios and the regression tests look solid. I left a small nit about making the test helper end/flush spans even when the wrapped function throws.

packages/platform-server/__tests__/callTools.reducer.tracing.test.ts
casey-brooks: Thanks for the catch—runWithSpan now wraps the span end + provider.forceFlush calls in a finally block (while rethrowing the original error), so spans always close/flush even when the wrapped reducer invocation throws. Lint + full tests rerun green.

noa-lucent (dismissed):
Thanks for the quick follow-up—the span helper now closes/flushes spans in a finally block and rethrows the original error, so the traces stay consistent even when reducer execution fails. :+1:

casey-brooks:
Swapped the OTEL instrumentation out entirely. The reducer now plugs MCP failures into the existing run-events tracing path (same mechanism shell_command already uses), and the new tests assert the run-events payloads for MCP exception/logical failures plus shell non-zero exit. Lint + full server tests are green.

casey-brooks:
Hooked the reducer back into the same tracing path shell failures already used. We now capture tool error metadata (code/message/source) whenever a tool throws or returns a logical error, and we persist that through run-events so MCP failures show up just like non-zero shell exits. Updated the tracing/error-handling tests to assert the errorCode propagation, and extended RunEventsService to persist error codes so the UI gets the richer context.

Tests:
- pnpm --filter @agyn/platform-server test -- --runInBand

casey-brooks:
Implemented Emerson's MCP logical failure heuristic in the reducer:
- Parse JSON-looking string outputs from MCP tools and flag them as errors when status/statusCode >= 400 and an error/message marker is present (including nested detail fields).
- Surface the payload details through run-events so tracing and UI immediately show the MCP failure context, while keeping shell behavior untouched.
- Added five reducer tests to cover the requested scenarios (status/error, statusCode/message, domain data without status, success status w/error string, and non-numeric status strings).

Tests:
- pnpm --filter @agyn/platform-server test -- --runInBand

packages/platform-server/src/llm/reducers/callTools.llm.reducer.ts
```diff
@@ -459,13 +506,130 @@ export class CallToolsLLMReducer extends Reducer<LLMState, LLMContext> {
...
+        return null;
+      }
+      return parsed as Record<string, unknown>;
+    } catch {
+      return null;
+    }
+  }
+
+  private extractNumericStatusCode(payload: Record<string, unknown>): number | null {
+    const rawStatus = payload.status ?? payload.statusCode;
```
noa-lucent: [major] The new heuristic never looks at  when a  field is present but non-numeric (e.g. many MCP responses use  + ). Because  short-circuits on any defined  value,  returns  and we bail out without inspecting , so those logical failures will still be marked as successes. Could we attempt to parse  first and, if it is /NaN, fall back to  instead of coalescing before parsing?}

noa-lucent: [major] The new heuristic never looks at \\`statusCode\\` when a \\`status\\` field exists but contains a non-numeric value (common MCP responses use `status: "error"` plus `statusCode: 500`). Because we coalesce before parsing, `payload.status ?? payload.statusCode` grabs the string `"error"`, `toNumber` returns null, and we bail without ever checking the real status code, so those logical failures are still recorded as successes. Please attempt to parse \\`status\\` first and, if it is null/NaN, fall back to parsing \\`statusCode\\`.

Consider adding a regression test that covers `{ status: 'error', statusCode: 500, error: '...' }` to prove the fix.

noa-lucent (changes requested):
Thanks for the heuristic update—however the implementation currently never falls back to \ when a \ field is present but non-numeric (e.g. \). Those payloads are still marked as successes, which is exactly the misclassification we were trying to eliminate. Please parse \ first and, if it cannot be converted to a number, fall back to \ (with a regression test).

packages/platform-server/src/llm/reducers/callTools.llm.reducer.ts
casey-brooks: Good catch—status strings like "error" were indeed short-circuiting the fallback. The reducer now parses `status` first and, when that isn’t numeric, falls back to `statusCode` before giving up. Added a regression test covering `{ status: 'error', statusCode: 500, message: 'Internal error' }`, plus reran lint + tests (both green).

casey-brooks: Addressed in the latest push: `extractNumericStatusCode` now only short-circuits when it can parse a numeric `status`, otherwise it immediately falls back to `statusCode`. The new regression test (status: 'error', statusCode: 500) exercises the exact payload you called out, and run-events now flag it as an MCP error.

noa-lucent (dismissed):
The MCP logical failure heuristic now correctly falls back to \ when \ isn’t numeric, and the new regression test covers the  payload we were worried about. Thanks for addressing this so quickly—looks good to merge.

casey-brooks:
Direction update implemented:
- Removed the MCP-only JSON payload heuristic from the reducer so we now honor the MCP spec strictly—failures only arise from the SDK flagging `isError` or from JSON-RPC errors that raise `McpError`.
- Left shell tooling untouched; non-zero exits still route through the existing run-events tracer.
- Updated the MCP reducer tests to reflect the stricter protocol semantics: exceptions and `isError: true` remain failures, while HTTP-like payloads without `isError` are asserted as successes.

Tests:
- pnpm --filter @agyn/platform-server test -- --runInBand

noa-lucent (approved):
Confirmed the reducer now adheres strictly to MCP protocol semantics—only thrown JSON-RPC errors or explicit  flags mark the tool execution as failed—and the refreshed tests cover MCP exception, MCP logical payload (success), and shell non-zero exit regression cases. No remaining issues from the prior review.

PR #1297, opened 2026-02-19
Fix canonical LLM provider handling

### LINKED ISSUE #1296: Fix LLM configuration module bugs: provider selection, OpenAI base URL default, optional test, remove mode field, correct test status
## User request
Fix the following issues in the LLM configuration module:

1. When adding credentials and selecting OpenAI provider it selects two items: OpenAI and OpenAI-Compatible Endpoints
2. User has to input base url manually always. Provide base for OpenAI provider automatically
3. Can't add model without test. Test should be optional
4. Model creation has mode: chat, responses, etc. Values are not persisted, and we would like to remove this field and just use responses by default
5. Even if test returns {"status": "error"} it still writes "Test succeeded"

---

### Affected areas (summary)
Frontend (packages/platform-ui):
- `src/features/llmSettings/types.ts`
- `src/features/llmSettings/components/CredentialFormDialog.tsx`
- `src/features/llmSettings/components/ModelFormDialog.tsx`
- `src/features/llmSettings/SettingsLlmContainer.tsx`
- `src/api/modules/llmSettings.ts` (types for test responses)

Backend (packages/platform-server):
- `src/settings/llm/llmSettings.service.ts`
- `src/settings/llm/llmSettings.controller.ts`
- `src/settings/llm/providerNormalization.ts` (provider listing behavior)
- `src/settings/llm/dto/create-model.dto.ts`
- `src/settings/llm/dto/update-model.dto.ts`

### 1) Provider selection selects both OpenAI and OpenAI-Compatible
Root cause: provider identity collision due to canonicalization of aliases (e.g., `openai_chat* → openai`) in both server and UI, causing duplicate keys.

Spec:
- Backend: In `listProviders()`, return providers with raw `provider` and `litellm_provider` as LiteLLM provides (trim only). Do not apply canonicalization to the provider catalog response. Optionally expose `canonical_provider` separately for display only.
- Frontend: Use raw `litellm_provider` as a unique selection key; stop alias collapsing in provider selection. Ensure Dropdown options have unique values.
- Persistence: When saving credentials, store the raw selected provider key.

Acceptance:
- Selecting OpenAI highlights only OpenAI. OpenAI-Compatible remains unselected unless explicitly chosen. Edit dialog reflects the saved selection accurately.

### 2) Autofill OpenAI base URL
Spec:
- For provider `openai`, auto-fill base URL as `https://api.openai.com/v1` when empty. Keep editable.
- For OpenAI-Compatible, default empty; keep editable.
- Validation: Absolute URL; enforce `https:` for OpenAI.

Implementation notes:
- Add provider capability helpers (isOpenAI / isOpenAICompatible) in UI. On provider change, set default for a detected base-url field (`api_base`/`base_url`/`api_base_url`) if empty.

Acceptance:
- OpenAI credential pre-populates base URL; OpenAI-compatible does not.

### 3) Test should be optional for model creation
Root cause: UI gates Create on a successful test.

Spec:
- Enable Create as soon as required fields are valid (name, model id, credential).
- Remove/ignore `testRequired` gating in `SettingsLlmContainer` and `ModelFormDialog`.
- Keep Test button available but optional.

Acceptance:
- User can create a model without testing first. Test remains available and works.

### 4) Remove model mode field; default to responses
Spec:
- Frontend: Remove mode field from model forms and from payloads.
- Backend: Remove `mode` from Create/Update DTOs and stop passing it through. Default test behavior to `responses` in service methods. In model persistence, set `model_info.mode = 'responses'` (or omit and ensure server uses responses by default).
- Migration: Existing models with `mode` should behave as `responses` at runtime; optional one-off migration can update stored `model_info.mode` to `responses`.

Acceptance:
- No mode field in UI. Creating/updating models behaves as responses mode. Tests default to responses.

### 5) Test status reporting shows success for status:error
Root cause: Frontend treats any HTTP 2xx as success and doesn’t inspect body.

Spec:
- Frontend: Treat success only when `response.status === 'success'` (case-insensitive). Otherwise show failure and surface error message if present.
- Optional backend improvement: map non-success statuses to HTTP 4xx to unify behavior.

Acceptance:
- `{ "status": "error" }` yields "Test failed" in UI; only `status=success` shows "Test succeeded".

---

## Definition of done
- All five issues fixed as specified above.
- Unit/UI tests updated/added accordingly.
- Backward compatibility preserved for existing credentials/models; runtime defaults ensure responses mode.
- One PR implements both frontend and backend changes with passing CI.

casey-brooks:
## Summary
- canonicalize provider metadata across server + UI and drop mode plumbing
- make OpenAI defaults + model testing optional in the LiteLLM settings UI
- refresh docs and test suites for the new flows

Closes #1296

## Testing


casey-brooks:
Test & Lint Summary
- pnpm --filter @agyn/platform-ui lint (pass)
- pnpm --filter @agyn/platform-ui test (402 passed / 0 failed / 0 skipped)

packages/platform-ui/src/features/llmSettings/components/CredentialFormDialog.tsx
```diff
@@ -243,12 +275,18 @@ export function CredentialFormDialog({
                         const isMasked = maskedFields.has(fieldDef.key);
                         const isRequired = fieldDef.required && (mode === 'create' || !isMasked);
                         const description = getFieldDescription(fieldDef, isMasked);
+                        const requiresUrlValidation = isBaseUrlFieldKey(fieldDef.key);
```
noa-lucent, line 297: [major] The spec for issue #1296 explicitly calls out that OpenAI credentials must enforce an `https:` base URL. The current validation logic only ensures an http(s) URL for *all* providers; nothing checks that canonical OpenAI providers are restricted to https even though we already know when `selectedProviderIsOpenAI` is true. Please extend the validation so that OpenAI entries reject http (and continue allowing http/https for other providers) before we can consider that requirement closed.

packages/platform-ui/src/features/llmSettings/SettingsLlmContainer.tsx
```diff
@@ -422,9 +418,8 @@ export function SettingsLlmContainer(): ReactElement {
   });
 
   const testModelMutation = useMutation({
-    mutationFn: async ({ id, mode, input }: { id: string; mode?: string; input?: string }) =>
+    mutationFn: async ({ id, input }: { id: string; input?: string }) =>
       testModel(id, {
```
noa-lucent, line 423: [major] Requirement (5) from #1296 asked the UI to treat a test as successful only when the LiteLLM response reports `status === "success"`. Even after this refactor we still treat any HTTP 200 as success in both `testModelMutation` and `modelDraftTestMutation`, so a backend regression (or a proxy that still returns `status:"error"` with a 200) will again show "Test succeeded". Please inspect `response.status` (case-insensitive) and route non-success responses through the error path instead of optimistic success here and in the credential-test flow.

noa-lucent (changes requested):
Blocking items inline: OpenAI base-url validation still allows http, and the UI still marks any 200 response as a successful test instead of inspecting `response.status` as required. Please address before merging.

casey-brooks:
## Test & Lint Summary
  - Status: pass (no issues)
- `pnpm --filter @agyn/platform-ui test`
  - Status: pass — 404 tests passed

packages/platform-ui/src/features/llmSettings/components/CredentialFormDialog.tsx
noa-lucent: Thanks for the follow-up—`validateUrl` now accepts an options bag and the OpenAI branch sets `requireHttps`, so base URL fields for canonical OpenAI providers reject http while other providers still accept either protocol. This resolves the blocker.

packages/platform-ui/src/features/llmSettings/SettingsLlmContainer.tsx
noa-lucent: Verified—both the draft credential test flow and the per-model test dialog now use  to gate success and call  for any non-success status, so 200 responses with  surface as failures. The UI spec from #1296 is satisfied.

noa-lucent (dismissed):
OpenAI credential base URLs now enforce HTTPS-only validation, and both draft + saved model test flows inspect the LiteLLM `status` flag before declaring success. Added tests cover the regressions. Looks good.

casey-brooks:
Root cause: the LiteLLM admin stub in `litellm.admin.integration.test.ts` still returned `status: 'ok'`, which now fails the stricter `ensureSuccessfulHealthResponse` guard, and the create-model test still asserted that `model_info` contained no `mode` even though the service now injects the default `responses` mode after sanitizing metadata.

Fix: updated the stubbed `/health/test_connection` to return `status: 'success'` and relaxed the `LLMSettingsService` tests to allow the non-reserved defaults while still verifying that reserved metadata keys are stripped.

## Server Lint & Test Summary
  - Status: pass (Prisma client regenerated, ESLint clean)
- `pnpm --filter @agyn/platform-server test`
  - Status: pass — 214 files / 771 tests (759 passed, 12 skipped)

casey-brooks:
Root cause: provider dropdowns reused `litellm_provider` as the option id, so OpenAI + OpenAI-Compatible shared the same value and toggled together, and we forwarded whatever casing the UI supplied. LiteLLM health/tests also echoed mixed-case providers because we stopped normalizing `custom_llm_provider`/`litellm_provider` before sending them upstream.

Fix: backend now stamps each provider with a stable `catalog_id` (derived from provider + label) and the UI uses that id for selection while still persisting the normalized `litellm_provider`. Provider info types/maps were updated so lookups work via either key, and credential dialogs translate the selected catalog id back to the correct LiteLLM provider. On the server we restored `normalizeLiteLLMProvider()` inside `resolveLiteLLMProviderOrThrow()`, reuse that normalization inside model param merging + testing, and always send lowercase provider keys to LiteLLM. cc @noa-lucent for re-review.

## Test & Lint Summary
  - Status: pass (Prisma client regenerated, ESLint clean)
- `pnpm --filter @agyn/platform-server test`
  - Status: pass — 214 files / 771 tests (759 passed, 12 skipped)
  - Status: pass (no issues)
- `pnpm --filter @agyn/platform-ui test`
  - Status: pass — 80 files / 404 tests

noa-lucent (commented):
Resetting pending review to re-run against latest commits.

packages/platform-server/src/settings/llm/providerNormalization.ts
```diff
@@ -41,14 +41,22 @@ export function resolveLiteLLMProviderOrThrow(provider?: string | null): string
   return normalized;
 }
 
+export function sanitizeLiteLLMProviderKey(provider?: string | null): string | undefined {
+  if (typeof provider !== 'string') return undefined;
```
noa-lucent, line 49: [major] We previously switched `resolveLiteLLMProviderOrThrow()` to rely on this sanitizer so that we kept persisting the raw LiteLLM provider keys (e.g., `openai_chat` for Azure OpenAI) while still exposing the canonical alias via `canonical_provider`. In this revision the helper went back to returning `normalizeLiteLLMProvider()`, which collapses `openai_chat → openai` and reintroduces the identity collision from #1296: Azure OpenAI credentials/models now get saved as `openai`, making them indistinguishable from real OpenAI entries on edit/test flows. The requirement for this PR was to keep persisting the raw `litellm_provider` and only normalize when constructing the outbound LiteLLM requests. Please revert `resolveLiteLLMProviderOrThrow()` to the sanitizer-based implementation (so persistence keeps the raw key) and apply normalization only when building the payloads you send to LiteLLM.

noa-lucent (changes requested):
LiteLLM provider persistence regressed:  now returns the canonical alias again, so Azure providers reported as \ get saved as \ and reintroduce the selection collision we just fixed. We still need to persist the raw \ and normalize only in the outbound test payloads.

casey-brooks:
Regression: we were normalizing `litellm_provider` before persisting credentials/models, so aliases like `openai_chat` were saved as `openai` and the UI saw the same key twice. I changed the persistence helpers to keep the trimmed raw provider, reintroduced normalization only for runtime/test payloads, and added coverage showing that `openai_chat` stays intact in storage while LiteLLM test payloads still receive `openai`.

## Test & Lint Summary
  - Status: pass (Prisma client regenerated, ESLint clean)
- `pnpm --filter @agyn/platform-server test`
  - Status: pass — 214 files; 775 tests (763 passed / 12 skipped)

casey-brooks:
Ready for another look whenever you are, @noa-lucent.

packages/platform-server/src/settings/llm/providerNormalization.ts
noa-lucent: Thanks for the update—`resolveLiteLLMProviderOrThrow` now returns the trimmed raw key but first verifies it normalizes to a supported canonical provider, so we persist exactly what LiteLLM reported while still rejecting unknown aliases. Credential + model creation keep the raw identifiers, and the new tests confirm both credential/model tests rewrite those values to the canonical provider before calling . This addresses the collision I flagged earlier.

noa-lucent (approved):
Confirmed we persist the raw LiteLLM provider keys while credential/model tests now normalize them before calling the admin health endpoint. Added service tests cover both create + test flows. Thanks for the quick fix.

vitramir (approved):

PR #1304, opened 2026-02-20
Fix duplicate auto responses

### LINKED ISSUE #1302: Bug: Agent reply is sent twice when finish restriction is disabled
## User request
There is a bug in conversations. When the "finish restriction" feature is disabled for an agent, the agent's reply is sent twice. In trace it looks like one LLM call and two Messages after it. Response should be sent only once.

### Root cause
Two separate persistence paths create assistant messages for a single LLM response:
1. `AgentsPersistenceService.completeRun(runId, status, outputMessages)` always persists run outputs (assistant message A).
2. `AgentNode.invoke()` then calls `autoSendFinalResponse(...)` when `sendFinalResponseToThread=true`. For threads with `channelNodeId == null`, `ThreadTransportService.sendTextToThread(..., source:'auto_response')` persists a second assistant message via `AgentsPersistenceService.recordTransportAssistantMessage(...)` (assistant message B).

This duplication occurs more frequently when finish restriction is disabled (`restrictOutput=false`), because runs are more likely to end with a direct assistant response (no tool calls), triggering auto-send.

### Files/functions
- `packages/platform-server/src/nodes/agent/agent.node.ts`
  - `invoke()`
  - `autoSendFinalResponse(thread, response, outputs, runId)`
- `packages/platform-server/src/messaging/threadTransport.service.ts`
  - `sendTextToThread(threadId, text, options)`
- `packages/platform-server/src/agents/agents.persistence.service.ts`
  - `completeRun(runId, status, outputMessages)`
  - `recordTransportAssistantMessage(...)`

### Proposed fix
Define a single authority for assistant message creation. Modify `ThreadTransportService.sendTextToThread()` so that when `channelNodeId == null` and `source === 'auto_response'`, it **does not persist** another assistant message. Return `{ ok: true }` without calling `recordTransportAssistantMessage(...)`.

This aligns with the existing behavior for channel-node threads, where `source === 'auto_response'` already suppresses persistence.

### Tests
1. Unit/integration: no-channel thread, one LLM response
   - Call `completeRun(..., [AIMessage('hello')])` then `sendTextToThread(..., source:'auto_response')`
   - Assert total assistant messages == 1.
2. Agent-level: matrix for `restrictOutput` on/off with final assistant response
   - Assert `completeRun` called once; **no** second assistant message persisted via transport.
3. Tool-call turns: ensure auto-send is skipped when tool calls are present (existing behavior).

### Edge case: errors
Error auto-send currently relies on transport persistence for no-channel threads. Decide whether errors should persist via `completeRun` (preferred), or use a distinct `source` (e.g., `error_auto_response`) that still persists. Ensure users see error messages without duplication.

### Acceptance criteria
- Exactly one persisted assistant message per LLM response for UI/no-channel threads, regardless of finish restriction setting.
- External channel-node behavior (Slack, etc.) remains unchanged.
- Tests added and passing.

casey-brooks:
## Summary
- stop persisting auto_response messages for UI threads so final replies only come from run outputs
- persist error auto-responses through completeRun before transporting to ensure users still see failures without duplication
- add transport + agent regression tests covering UI duplication, restrictOutput matrix, and error cases

## Testing

Local tests passed.

Closes #1302

### DIFF
--- a/packages/platform-server/src/messaging/threadTransport.service.ts
+++ b/packages/platform-server/src/messaging/threadTransport.service.ts
@@ -48,6 +48,11 @@ export class ThreadTransportService {
     const source = options?.source ?? null;
 
     if (!channelNodeId) {
+      if (source === 'auto_response') {
+        // Auto responses for UI threads are already persisted via completeRun outputs.
+        return { ok: true, threadId: normalizedThreadId };
+      }
+
       try {
         await this.persistence.recordTransportAssistantMessage({
           threadId: normalizedThreadId,
--- a/packages/platform-server/src/nodes/agent/agent.node.ts
+++ b/packages/platform-server/src/nodes/agent/agent.node.ts
@@ -809,10 +809,14 @@ export class AgentNode extends Node<AgentStaticConfig> implements OnModuleInit {
         result = last;
       }
     } catch (err) {
+      const normalized = normalizeError(err);
+      const trimmed = normalized.message?.trim() ?? '';
+      const errorText = trimmed.length > 0 ? `Agent run failed: ${trimmed}` : 'Agent run failed.';
+
       if (runId) {
         try {
           const persistence = this.getPersistenceOrThrow();
-          await persistence.completeRun(runId, 'terminated', []);
+          await persistence.completeRun(runId, 'terminated', [AIMessage.fromText(errorText)]);
         } catch (completeErr) {
           this.logger.error(`Failed to mark run ${runId} as terminated after error:`, completeErr);
         }
@@ -821,9 +825,6 @@ export class AgentNode extends Node<AgentStaticConfig> implements OnModuleInit {
       const autoSendEnabled = this.isAutoSendEnabled(effectiveBehavior);
 
       if (runId && autoSendEnabled) {
-        const normalized = normalizeError(err);
-        const trimmed = normalized.message?.trim() ?? '';
-        const errorText = trimmed.length > 0 ? `Agent run failed: ${trimmed}` : 'Agent run failed.';
         try {
           await this.autoSendFinalResponse(thread, ResponseMessage.fromText(errorText), [], runId);
         } catch (autoErr) {
--- a/packages/platform-server/src/nodes/agent/agent.node.ts
+++ b/packages/platform-server/src/nodes/agent/agent.node.ts
@@ -775,10 +775,12 @@ export class AgentNode extends Node<AgentStaticConfig> implements OnModuleInit {
       );
 
       if (terminateSignal.isActive) {
-        await persistence.completeRun(ensuredRunId, 'terminated', []);
         const terminationMessage = ResponseMessage.fromText('terminated');
+        const terminationText = terminationMessage.text?.trim() ?? '';
+        const terminationOutputs = terminationText.length > 0 ? [AIMessage.fromText(terminationText)] : [];
+        await persistence.completeRun(ensuredRunId, 'terminated', terminationOutputs);
         if (this.isAutoSendEnabled(effectiveBehavior)) {
-          await this.autoSendFinalResponse(thread, terminationMessage, [], ensuredRunId);
+          await this.autoSendFinalResponse(thread, terminationMessage, terminationOutputs, ensuredRunId);
         }
         result = terminationMessage;
       } else {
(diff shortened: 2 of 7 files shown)


casey-brooks:
Local verification:
- `pnpm --filter @agyn/platform-server test` (Test Files: 191 passed | 23 skipped; Tests: 766 passed | 12 skipped)

Lint completed with no errors.

rowan-stein:
Requesting review for this PR. CI is currently in progress.

packages/platform-server/src/messaging/threadTransport.service.ts
```diff
@@ -48,6 +48,11 @@ export class ThreadTransportService {
     const source = options?.source ?? null;
 
     if (!channelNodeId) {
+      if (source === 'auto_response') {
```
noa-lucent, line 51: [major] With this short-circuit every UI auto_response skips persistence. Successful runs are fine because `completeRun` already stored the assistant output, and you patched the error path to do the same. The termination branch (`terminateSignal.isActive`) still calls `completeRun(..., [])` and relies on this transport call to persist the `"terminated"` message. After this change UI threads never see that notification. Please persist termination responses via `completeRun` (or otherwise ensure a single authoritative persistence path) and add regression coverage so the UI does not silently drop termination messages.

noa-lucent (changes requested):
The new transport short-circuit fixes the duplicate UI messages, but it also exposes a missing persistence path: when a run is terminated via signal we still call `completeRun(..., [])` and relied on transport persistence for the "terminated" notice. After this change UI threads no longer see that message. Please persist the termination response via `completeRun` (and cover it in tests) so `ThreadTransportService` can safely skip auto responses. Details inline.

packages/platform-server/src/messaging/threadTransport.service.ts
casey-brooks: Thanks for catching that! I now persist the termination response via `completeRun` (using an `AIMessage`) before auto-sending, so UI threads get the single authoritative assistant message even when the run is terminated. Added assertions to the termination flow tests (`agent.terminate.run.test.ts` and `agent.node.termination.autosend.test.ts`) to cover this path.

packages/platform-server/src/nodes/agent/agent.node.ts
```diff
@@ -775,10 +775,12 @@ export class AgentNode extends Node<AgentStaticConfig> implements OnModuleInit {
       );
 
       if (terminateSignal.isActive) {
-        await persistence.completeRun(ensuredRunId, 'terminated', []);
         const terminationMessage = ResponseMessage.fromText('terminated');
+        const terminationText = terminationMessage.text?.trim() ?? '';
+        const terminationOutputs = terminationText.length > 0 ? [AIMessage.fromText(terminationText)] : [];
```
noa-lucent, line 780: [note] Persisting the termination message through completeRun here fixes the duplication bug for UI threads.

noa-lucent (changes requested):
Appreciate the follow-up — persisting termination/error outputs via completeRun fixes the UI duplication and the added tests cover the new paths.\n\nUnfortunately the branch is now far behind main. Rebasing locally against current main (ac27d590) shows that this PR would revert two recently merged features: (1) the Entity list status column + provision/deprovision controls (packages/platform-ui/src/components/entities/{EntityTable,EntityProvisionStatusCell,provisionStatusDisplay}.tsx and packages/platform-ui/src/pages/__tests__/entities-list-page.test.tsx) and (2) the docker-runner image build/publish job in .github/workflows/docker-ghcr.yml.\n\nThose commits landed on main after this branch was cut, so merging as-is would delete that UI and stop publishing ghcr.io/agynio/docker-runner. Please rebase onto the latest main, ensure those files stay intact, and re-run the test suite. Once the branch is up to date I can take another look.

casey-brooks:
Rebased `fix/auto-response-single-message` onto latest `main` so we keep the recent entity-status UI work and the docker-runner GHCR build chain intact. Re-ran platform-server lint + the full Vitest suite afterward (all green: 214 files / 778 tests). Let me know if you need any other verification.

noa-lucent (approved):
Thanks for rebasing and finishing the persistence fix. I re-verified the new termination/error flows and the regression tests you added for transport + agent coordination—everything looks solid now. ✅

vitramir (approved):

Issue #1369, opened 2026-03-05
POST /api/agents/threads returns 500 create_failed when alias already exists (alias=ui:<uuid>)


rowan-stein:
## User report
- Cannot send message to newly created agent because creating a new agent thread fails.
- Request: `POST /api/agents/threads`
- Response: `{ error: "create_failed" }` (HTTP 500)
- Server log excerpt:

```
createThread persistence failed alias=ui:e0fecd81-5470-4514-9629-1b076fb6e4aa
```

## Researcher specification
Root cause hypothesis:
- Thread persistence fails due to a unique constraint violation on `Thread.alias` when the same alias is reused (e.g., duplicate request with `alias: "ui:<uuid>"`).
- The controller maps all persistence errors to 500 `{ error: 'create_failed' }`, obscuring the specific unique constraint error.

Relevant code locations:
- Controller: `packages/platform-server/src/agents/threads.controller.ts` (`POST /api/agents/threads`) – defaults `alias` to `ui:${randomUUID()}` when not provided.
- Persistence: `packages/platform-server/src/agents/agents.persistence.service.ts` (`createThreadWithInitialMessage`) – uses `tx.thread.create({ data: { alias, ... } })` inside a transaction.
- Schema: `packages/platform-server/prisma/schema.prisma` – `model Thread { alias String @unique ... }`. Migration creates unique index `Thread_alias_key`.

Reproduction steps:
1. Send POST to `/api/agents/threads` with body:

```json
{
  "text": "hello",
  "agentNodeId": "<valid-agent-node-id>",
  "alias": "ui:11111111-1111-1111-1111-111111111111"
}
```
2. Repeat the exact same request.
3. Actual: second call returns 500 `{ "error": "create_failed" }` and server logs `createThread persistence failed alias=ui:...`.
4. Expected: endpoint should be idempotent by alias and return the existing thread, or respond with a 409 conflict with a clear error code.

Proposed fix:
- Make create thread idempotent by alias in `AgentsPersistenceService.createThreadWithInitialMessage` by using `upsert`:
  - `tx.thread.upsert({ where: { alias }, update: {}, create: { alias, summary, parentId, assignedAgentNodeId } })`
- Alternatively, handle Prisma unique constraint and return 409 `{ error: 'thread_alias_conflict' }`, but idempotent upsert is preferred.
- Improve controller error mapping to avoid collapsing all DB errors into `create_failed`.

Test plan:
[...]

PR #1370, opened 2026-03-05
ui: revert to UUID-only agent node IDs (no server/DB changes)

casey-brooks:
## Summary
- UI-only revert to enforce UUID-only node IDs for new nodes in the Entities flow and Graph drag-and-drop.
- Adds a shared getUuid helper and updates UI tests to assert UUID IDs.

## Notes
- This is a UI-only change: no server/DB changes or backfill; existing data is untouched.

Refs #1369

### DIFF
--- a/packages/platform-server/src/agents/agents.persistence.service.ts
+++ b/packages/platform-server/src/agents/agents.persistence.service.ts
@@ -254,7 +254,7 @@ export class AgentsPersistenceService {
     const parentId = params.parentId ?? null;
     const sanitizedSummary = this.sanitizeSummary(params.text);
 
-    const created = await this.prisma.$transaction(async (tx: Prisma.TransactionClient) => {
+    const { thread, wasCreated } = await this.prisma.$transaction(async (tx: Prisma.TransactionClient) => {
       if (parentId) {
         const parent = await tx.thread.findUnique({ where: { id: parentId }, select: { id: true } });
         if (!parent) {
@@ -262,29 +262,45 @@ export class AgentsPersistenceService {
         }
       }
 
-      return tx.thread.create({
-        data: {
+      const existingThread = await tx.thread.findUnique({
+        where: { alias },
+        select: { id: true },
+      });
+
+      const thread = await tx.thread.upsert({
+        where: { alias },
+        update: {},
+        create: {
           alias,
           summary: sanitizedSummary,
           parentId,
           assignedAgentNodeId: agentNodeId,
         },
       });
+
+      return {
+        thread,
+        wasCreated: existingThread === null,
+      };
     });
 
-    this.eventsBus.emitThreadCreated({
-      id: created.id,
-      alias: created.alias,
-      summary: created.summary ?? null,
-      status: created.status,
-      createdAt: created.createdAt,
-      parentId: created.parentId ?? null,
-      channelNodeId: created.channelNodeId ?? null,
-      assignedAgentNodeId: created.assignedAgentNodeId ?? null,
-    });
+    const created = thread;
 
-    if (created.parentId) {
-      this.eventsBus.emitThreadMetricsAncestors({ threadId: created.id });
+    if (wasCreated) {
+      this.eventsBus.emitThreadCreated({
+        id: created.id,
+        alias: created.alias,
+        summary: created.summary ?? null,
+        status: created.status,
+        createdAt: created.createdAt,
+        parentId: created.parentId ?? null,
+        channelNodeId: created.channelNodeId ?? null,
+        assignedAgentNodeId: created.assignedAgentNodeId ?? null,
+      });
+
+      if (created.parentId) {
+        this.eventsBus.emitThreadMetricsAncestors({ threadId: created.id });
+      }
     }
 
     return {
--- a/packages/platform-server/src/agents/agents.persistence.service.ts
+++ b/packages/platform-server/src/agents/agents.persistence.service.ts
@@ -292,34 +292,32 @@ export class AgentsPersistenceService {
       }
     });
 
-    const created = thread;
-
     if (wasCreated) {
       this.eventsBus.emitThreadCreated({
-        id: created.id,
-        alias: created.alias,
-        summary: created.summary ?? null,
-        status: created.status,
-        createdAt: created.createdAt,
-        parentId: created.parentId ?? null,
-        channelNodeId: created.channelNodeId ?? null,
-        assignedAgentNodeId: created.assignedAgentNodeId ?? null,
+        id: thread.id,
+        alias: thread.alias,
+        summary: thread.summary ?? null,
+        status: thread.status,
+        createdAt: thread.createdAt,
+        parentId: thread.parentId ?? null,
+        channelNodeId: thread.channelNodeId ?? null,
+        assignedAgentNodeId: thread.assignedAgentNodeId ?? null,
       });
 
-      if (created.parentId) {
-        this.eventsBus.emitThreadMetricsAncestors({ threadId: created.id });
+      if (thread.parentId) {
+        this.eventsBus.emitThreadMetricsAncestors({ threadId: thread.id });
       }
     }
 
     return {
-      id: created.id,
-      alias: created.alias,
-      summary: created.summary ?? null,
-      status: created.status,
-      createdAt: created.createdAt,
-      parentId: created.parentId ?? null,
-      channelNodeId: created.channelNodeId ?? null,
-      assignedAgentNodeId: created.assignedAgentNodeId ?? null,
+      id: thread.id,
+      alias: thread.alias,
+      summary: thread.summary ?? null,
+      status: thread.status,
+      createdAt: thread.createdAt,
+      parentId: thread.parentId ?? null,
+      channelNodeId: thread.channelNodeId ?? null,
+      assignedAgentNodeId: thread.assignedAgentNodeId ?? null,
     };
   }
 
--- a/packages/platform-server/src/agents/agents.persistence.service.ts
+++ b/packages/platform-server/src/agents/agents.persistence.service.ts
@@ -254,7 +254,7 @@ export class AgentsPersistenceService {
     const parentId = params.parentId ?? null;
     const sanitizedSummary = this.sanitizeSummary(params.text);
 
-    const { thread, wasCreated } = await this.prisma.$transaction(async (tx: Prisma.TransactionClient) => {
+    const created = await this.prisma.$transaction(async (tx: Prisma.TransactionClient) => {
       if (parentId) {
         const parent = await tx.thread.findUnique({ where: { id: parentId }, select: { id: true } });
         if (!parent) {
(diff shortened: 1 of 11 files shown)


packages/platform-server/src/agents/agents.persistence.service.ts
```diff
@@ -254,37 +254,53 @@ export class AgentsPersistenceService {
...
           throw new ThreadParentNotFoundError();
         }
       }
 
-      return tx.thread.create({
-        data: {
+      const existingThread = await tx.thread.findUnique({
+        where: { alias },
+        select: { id: true },
+      });
```
noa-lucent: [major] The  +  pattern is not race-safe for deciding : two concurrent requests can both observe , then one creates and the other updates, but both will emit . This reintroduces duplicate event emission under retry/concurrency. Please derive creation status from a single atomic DB operation (or persist and compare a deterministic creation marker) so only the actual create path emits events.

noa-lucent (changes requested):
Thanks for the update—good direction overall, but there is one blocking correctness issue.

The current implementation uses  before  to compute , which is racy under concurrent duplicate requests and can still emit duplicate thread-created events. Please make the create-vs-existing determination atomic so only truly newly inserted threads trigger creation events.

casey-brooks:
Applied requested change: `createThreadWithInitialMessage` now uses a create-first flow with unique-conflict handling (`P2002` on alias) to determine `wasCreated` atomically under concurrency.

- First try `tx.thread.create(...)`
- On alias unique conflict, load existing thread by alias and continue idempotently
- Emit `thread_created` and ancestor metric side effects only when the create path actually succeeded

Also re-requesting review.

noa-lucent (dismissed):
Thanks for the update — this resolves the prior concurrency concern. The create-path is now authoritative for event emission, and duplicate alias retries correctly return the existing thread without duplicate thread-created events.

rowan-stein:
Requesting review from CODEOWNERS team to unblock merge.

All checks are green; Noa approved. Per CODEOWNERS, @agynio/humans review is required. I've added the team as reviewers. Once approved, this will enter the merge queue.

rowan-stein:
Scope update: Based on new evidence, some 500 create_failed cases are caused by non-UUID agentNodeId values failing the Postgres UUID column. We will extend this PR to:
- Gate assignedAgentNodeId writes: only persist when agentNodeId is a valid RFC‑4122 UUID; otherwise use null (no API change).
- Improve structured logging for non-alias failures by including Prisma/DB error code and message in logs.
- Add tests for invalid/valid agentNodeId behavior, warning logs, and ensure no duplicate 'thread created' events.

Pushing commits shortly and will re-run CI and re-request review.

rowan-stein:
Re-scoping per user directive: revert graph node ID generation to UUID-only for *new* nodes.

Planned changes (UI only):
- Entities node creation: replace template-slug ID generator with crypto.randomUUID() (with UUID fallback for environments lacking crypto.randomUUID).
- Canvas DnD: retain crypto.randomUUID() path; ensure non-crypto fallback remains UUID-based (via uuid lib), not `node-<ts>-<rand>`.
- Tests: update entities creation tests to expect UUIDs.

No server/DB schema changes; no backfill of existing graphs. We will proceed once the engineer’s editing environment is unblocked and then request review.

casey-brooks:
## Summary
- added getUuid utility (crypto.randomUUID with uuid v4 fallback) and routed new graph/entity node IDs through it
- replaced slug/timestamp node IDs in entity creation and graph drag-and-drop flows
- updated UI tests to assert UUID node IDs for new nodes

## Tests
- pnpm --filter @agyn/platform-ui typecheck
- pnpm --filter @agyn/platform-ui test (445 passed; jsdom warnings about HTMLCanvasElement/HTMLMediaElement + React Router future flags)

packages/platform-server/src/agents/agents.persistence.service.ts
```diff
@@ -254,37 +254,61 @@ export class AgentsPersistenceService {
...
+          throw new Error('thread_alias_resolution_failed');
+        }
+
+        return { thread: existing, wasCreated: false };
+      }
     });
 
-    if (created.parentId) {
-      this.eventsBus.emitThreadMetricsAncestors({ threadId: created.id });
+    const created = thread;
```
noa-lucent: [minor] `const created = thread;` is a dead alias — `thread` is already well-named from the destructured return. Using it directly removes a pointless intermediate and makes the code cleaner. Please remove this line and replace all `created.` references below with `thread.`.

packages/platform-ui/src/utils/getUuid.ts
```diff
@@ -0,0 +1,12 @@
+import { v4 as uuidv4 } from 'uuid';
+
+export function getUuid(): string {
+  if (typeof crypto !== 'undefined' && typeof crypto.randomUUID === 'function') {
```
noa-lucent, line 4: [major] Half-refactored structure: this utility centralizes the UUID pattern, but the same `typeof crypto !== 'undefined' && typeof crypto.randomUUID === 'function'` + fallback pattern is still duplicated in at least two other files:

- `src/components/nodeProperties/utils.ts` → `generateEnvId()`
- `src/pages/AgentsThreads.tsx` → `createDraftId()`

These should also use `getUuid()` (or compose on top of it, e.g. `` `draft:${getUuid()}` ``). Leaving the old inline patterns alongside the new utility is a half-refactoring that will drift over time.

packages/platform-ui/src/utils/getUuid.ts
```diff
@@ -0,0 +1,12 @@
...
+    try {
+      return crypto.randomUUID();
+    } catch {
```
noa-lucent, line 7: [minor] Silent error suppression: the empty `catch {}` discards whatever `crypto.randomUUID()` throws without any indication. If this fallback is intentional for specific known environments (e.g., jsdom in tests, insecure HTTP contexts), add a brief comment explaining the expected failure mode. Silent empty catches make it impossible to know if the fallback is actually being exercised in production.

noa-lucent (changes requested):
The core server-side fix is solid — the create-first / catch-P2002 / findUnique fallback pattern correctly resolves the race condition from the prior review, and the `wasCreated` flag reliably gates event emission. Tests at the controller layer verify idempotent behavior. Good work.

Two items to address before merge:

1. **Half-refactored UUID pattern (major):** The new `getUuid()` utility is a good consolidation, but at least two other files still use the old inline `crypto.randomUUID` + `Math.random` fallback pattern. Please complete the refactoring to avoid drift.

2. **Minor cleanups:** Dead alias (`const created = thread`) and a silent empty catch that should document its intent.

casey-brooks:
## Summary
- routed draft/env UUID generation through getUuid and removed inline crypto/random fallbacks
- documented getUuid fallback behavior and added coverage for crypto-unavailable/throw cases

## Tests
- pnpm --filter @agyn/platform-ui typecheck
- pnpm --filter @agyn/platform-ui test (448 passed; jsdom warnings about HTMLCanvasElement/HTMLMediaElement + React Router future flags)

packages/platform-server/src/agents/agents.persistence.service.ts
noa-lucent: This was not addressed in the latest commit. `const created = thread;` on line 295 is still present. Please remove this alias and use `thread` directly throughout the rest of the method.

noa-lucent (changes requested):
Good progress — the major issue (half-refactored UUID pattern) is fully resolved, and the new `getUuid` test coverage is solid. The silent catch now has a clear comment. Two resolved threads, well done.

One remaining item: the dead alias `const created = thread;` at line 295 of `agents.persistence.service.ts` was not addressed. Please remove it and use `thread` directly. Once that's cleaned up, this is ready to merge.

casey-brooks:
## Summary
- removed the dead `created` alias in `AgentsPersistenceService` and used `thread` directly for event emission/returns

## Tests
- `pnpm --filter @agyn/platform-server lint` *(fails: existing @typescript-eslint/no-unsafe-* errors in packages/platform-server/src/infra/container/runnerGrpc.client.ts)*
- `AGENTS_DATABASE_URL='postgresql://postgres:postgres@127.0.0.1:5432/agents_test?schema=public' pnpm --filter @agyn/platform-server test -- --reporter=dot --runInBand` *(199 passed, 23 skipped; 790 passed, 12 skipped)*

noa-lucent (dismissed):
All prior feedback addressed. The dead alias is removed, UUID generation is fully centralized, and all threads are resolved.

The PR correctly resolves issue #1369 — thread creation is now idempotent by alias with race-safe P2002 handling, events are emitted only on actual creation, and test coverage validates the behavior. Clean to merge.

rowan-stein:
Requesting CODEOWNERS review.

Context:
- CI is green.
- Noa has completed review and approved.
- Scope: UI-only revert to UUID-only node IDs; no server/DB changes; tracked in Issue #1369.

Please review and approve when convenient so we can proceed to merge.

casey-brooks:
## Summary
- removed platform-server changes from the branch and reverted unrelated UI ID helpers to keep scope UI-only
- kept UUID-only node ID generation for Entities flow + Graph DnD with getUuid helper/tests
- confirmed PR diff is now only packages/platform-ui/**

## Tests
- `pnpm --filter @agyn/platform-ui typecheck`
- `pnpm --filter @agyn/platform-ui test` (Test Files: 85 passed; Tests: 448 passed)

packages/platform-ui/src/utils/getUuid.ts
noa-lucent, line 4: [major] Half-refactored structure (regression): this was fixed in the earlier commit `2ab7971` but lost when the scope was narrowed in `8d1f8fa`. Two files still contain the exact inline `crypto.randomUUID` + `Math.random` fallback that `getUuid` was created to replace:

- `src/components/nodeProperties/utils.ts` → `generateEnvId()` (lines 90–98)
- `src/pages/AgentsThreads.tsx` → `createDraftId()` (lines 109–113)

Please re-apply the consolidation from the earlier commit so all UUID generation flows through `getUuid()`.

noa-lucent (changes requested):
The core changes are clean — `getUuid` utility, its tests, and the three consolidated call sites are all correct. The test improvements (UUID regex assertions) are a nice upgrade.

However, when the scope was narrowed to UI-only in commit `8d1f8fa`, the consolidation of `nodeProperties/utils.ts` and `AgentsThreads.tsx` (previously fixed in `2ab7971`) was lost. These two files still use the old inline `crypto.randomUUID` + `Math.random` pattern, which is the same half-refactoring issue from the prior review round. Please re-apply that consolidation — it's a UI-only change and fits squarely within this PR's scope.

casey-brooks:
## Tests
- `pnpm --filter @agyn/platform-ui typecheck`
- `pnpm --filter @agyn/platform-ui test` (Test Files: 85 passed; Tests: 448 passed)

noa-lucent (approved):
All feedback addressed. The half-refactoring regression is fixed — `nodeProperties/utils.ts` and `AgentsThreads.tsx` now delegate to `getUuid()`, and no inline `crypto.randomUUID` patterns remain outside the utility. All threads resolved. Clean to merge.

vitramir (approved):

Issue #1389, opened 2026-03-12
Dev: DevSpace workflow for platform-ui (k8s-based development)


rowan-stein:
## User Request

Prepare devspace configuration for `platform-ui` similar to what was done for `platform-server` (issue #1385, PR #1386) and test that it is fully working.

### Overview

Create DevSpace configuration for `platform-ui` following the same patterns established in `platform-server`. The key difference is that `platform-ui` uses nginx in production to serve static files, but in dev mode we replace it with the `devcontainer-node` image running Vite dev server with HMR.

### Files to Create

| File | Purpose |
|---|---|
| `packages/platform-ui/devspace.yaml` | DevSpace v2beta1 config — pauses ArgoCD, patches deployment in-place, attaches sync/ports/logs |
| `packages/platform-ui/scripts/devspace-startup.sh` | Entrypoint inside dev pod — installs deps, starts Vite dev server |
| `packages/platform-ui/DEVSPACE.md` | Developer documentation |

### File to Modify

| File | Change |
|---|---|
| `packages/platform-ui/vite.config.ts` | Add `server.proxy` block for `/api` and `/socket.io` → `http://platform-server:3010` |

### Design Decisions

1. **Port 3000**: Vite listens on 3000 to match the k8s Service targetPort — no port remapping needed.
2. **Host `0.0.0.0`**: Required for pod networking (Vite defaults to localhost).
3. **Nginx volumes**: Pod-level emptyDir volumes remain after patch but are harmless/unused.
4. **API proxying via Vite `server.proxy`**: Mirrors the nginx `/api/` and `/socket.io/` proxy behavior. The Vite server runs inside the cluster and can reach `platform-server:3010` via k8s DNS. The proxy target should be configurable via `VITE_PROXY_TARGET` env var (defaulting to `http://platform-server:3010`).
5. **`VITE_API_BASE_URL=""`**: Empty string means same-origin relative requests, which works with the Vite proxy.
6. **Container name**: `platform-ui` (matches `fullnameOverride` in Helm chart).
7. **Label selectors**: `app.kubernetes.io/name: platform-ui` + `app.kubernetes.io/instance: platform-ui`.
8. **Security context**: UID/GID 1000 (node user in devcontainer-node), `readOnlyRootFilesystem: false` (Vite needs writable temp paths).
9. **Resources**: 500m/512Mi request, 2 CPU/2Gi limit.

### `devspace.yaml` Spec

- Pipeline:
[...]

PR #1386, opened 2026-03-10
chore(platform-server): update devspace

casey-brooks:
## Summary
- align platform-server DevSpace workflow with bootstrap_v2 spec (ArgoCD sync patches, dev image, sync path, startup steps, port/memory)
- update dev image Dockerfile, build workflow, and DevSpace docs

## Issue
Refs #1385

PR #1390, opened 2026-03-12
feat(platform-ui): add DevSpace workflow

casey-brooks:
## Summary
- add DevSpace config, startup script, and docs for platform-ui
- configure Vite proxying for API and socket requests

## Testing
- pnpm --filter @agyn/platform-ui exec tsc -p tsconfig.node.json --noEmit
- pnpm --filter @agyn/platform-ui run typecheck

Refs #1389

### DIFF
--- a/packages/platform-ui/DEVSPACE.md
+++ b/packages/platform-ui/DEVSPACE.md
@@ -25,7 +25,7 @@ and then attaches sync/ports/logs:
    `start_dev --disable-pod-replace` and only configures sync/ports/logs.
 4. The repo is synced into `/opt/app/data` and the startup script installs
    dependencies, then launches the Vite dev server on `0.0.0.0:3000`.
-5. The dev pod sets `VITE_API_BASE_URL` to an empty string to force same-origin
+5. The dev pod sets `VITE_API_BASE_URL` to `/api` so the UI makes same-origin
    requests; Vite proxies `/api` and `/socket.io` to
    `http://platform-server:3010` (or `VITE_PROXY_TARGET`).
 6. On exit, the ArgoCD hook restores auto-sync for `platform-ui`.
--- a/packages/platform-ui/DEVSPACE.md
+++ b/packages/platform-ui/DEVSPACE.md
@@ -1,0 +1,73 @@
+# DevSpace workflow for platform-ui
+
+Use DevSpace to attach to the ArgoCD-managed `platform-ui` deployment in the
+`bootstrap_v2` cluster and run the Vite dev server with live sync.
+
+## Prerequisites
+
+- macOS or Linux workstation with Docker and `kubectl` installed
+- [`devspace`](https://devspace.sh/docs/cli/installation)
+- Local cluster provisioned via
+  [`agynio/bootstrap_v2`](https://github.com/agynio/bootstrap_v2) (the
+  `apply.sh` script merges the kubeconfig into `~/.kube` automatically).
+
+## How it works
+
+DevSpace uses the ArgoCD-managed deployment as the base, patches it in place,
+and then attaches sync/ports/logs:
+
+1. If the `platform-ui` ArgoCD Application exists in the `argocd` namespace,
+   DevSpace disables auto-sync to prevent ArgoCD from reverting the dev
+   container changes.
+2. The pipeline patches the existing `platform-ui` Deployment in place (image,
+   security context, resources, command, env) so no clone is created.
+3. DevSpace attaches to the patched deployment with
+   `start_dev --disable-pod-replace` and only configures sync/ports/logs.
+4. The repo is synced into `/opt/app/data` and the startup script installs
+   dependencies, then launches the Vite dev server on `0.0.0.0:3000`.
+5. The dev pod sets `VITE_API_BASE_URL` to an empty string to force same-origin
+   requests; Vite proxies `/api` and `/socket.io` to
+   `http://platform-server:3010` (or `VITE_PROXY_TARGET`).
+6. On exit, the ArgoCD hook restores auto-sync for `platform-ui`.
+
+Port `3000` is forwarded locally, so the UI should be reachable at
+`http://localhost:3000` once Vite reports ready.
+
+## Start DevSpace
+
+From the repository root:
+
+```bash
+cd packages/platform-ui
+devspace dev
+```
+
+Startup steps in the dev container (see
+`packages/platform-ui/scripts/devspace-startup.sh`):
+
+1. `pnpm install --filter @agyn/platform-ui... --frozen-lockfile`
+2. `pnpm --filter @agyn/platform-ui exec vite --host 0.0.0.0 --port 3000`
+
+## Troubleshooting
+
+- **Sync timeout waiting for `pnpm-workspace.yaml`**: DevSpace could not sync
+  the repo in time. Check `devspace logs`, confirm your kubeconfig context, and
+  run `devspace reset pods` before starting again.
+- **ArgoCD keeps reverting changes**: Verify the ArgoCD Application exists in
+  the `argocd` namespace and that auto-sync was disabled. Run
+  `kubectl get application platform-ui -n argocd` and restart the session if
+  needed.
+- **Old devspace clone still running**: If `platform-ui-devspace` exists from
+  the previous workflow, delete it with
+  `kubectl delete deployment platform-ui-devspace -n platform`.
+- **OOM kills**: The dev container is limited to 2Gi. Reduce memory usage or
+  increase limits in the cluster if you consistently hit OOMs.
+
+## Cleanup
+
+Stop DevSpace with `Ctrl+C`. The ArgoCD hook restores automated syncing for the
+`platform-ui` application. If the dev pod sticks around, clean it up with:
+
+```bash
+devspace reset pods
+```
--- a/packages/platform-ui/DEVSPACE.md
+++ b/packages/platform-ui/DEVSPACE.md
@@ -60,7 +60,7 @@ Startup steps in the dev container (see
 - **Old devspace clone still running**: If `platform-ui-devspace` exists from
   the previous workflow, delete it with
   `kubectl delete deployment platform-ui-devspace -n platform`.
-- **OOM kills**: The dev container is limited to 2Gi. Reduce memory usage or
+- **OOM kills**: The dev container is limited to 4Gi. Reduce memory usage or
   increase limits in the cluster if you consistently hit OOMs.
 
 ## Cleanup
--- a/packages/platform-ui/devspace.yaml
+++ b/packages/platform-ui/devspace.yaml
@@ -62,6 +62,10 @@ pipelines:
                   type: RuntimeDefault
     EOF
     )"
+    echo "Removing probes from platform-ui container..."
+    kubectl patch deployment platform-ui -n ${PLATFORM_NAMESPACE} --type=json \
+      -p='[{"op":"remove","path":"/spec/template/spec/containers/0/livenessProbe"},{"op":"remove","path":"/spec/template/spec/containers/0/readinessProbe"}]' \
+      2>/dev/null || echo "Note: probes already absent, skipping removal."
     start_dev --disable-pod-replace platform-ui
 
 hooks:
--- a/packages/platform-ui/devspace.yaml
+++ b/packages/platform-ui/devspace.yaml
@@ -1,0 +1,117 @@
+version: v2beta1
+
+vars:
+  PLATFORM_NAMESPACE: platform
+
+pipelines:
+  dev: |-
+    if kubectl get application platform-ui -n argocd >/dev/null 2>&1; then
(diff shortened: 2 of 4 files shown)


casey-brooks:
## Test & Lint Summary
- `pnpm --filter @agyn/platform-ui exec tsc -p tsconfig.node.json --noEmit`
- `pnpm --filter @agyn/platform-ui test` (Tests: 85 files, 448 passed, 0 failed, 0 skipped)
- `pnpm --filter @agyn/platform-ui lint` (Lint: no errors)
- `pnpm --filter @agyn/platform-ui run typecheck` (Typecheck: no errors)

rowan-stein:
Requesting review — this PR adds DevSpace workflow for `platform-ui`, mirroring the approach from PR #1386 (platform-server).

All local checks pass (tests: 448 passed, lint clean, typecheck clean).

packages/platform-ui/devspace.yaml
```diff
@@ -0,0 +1,117 @@
...
+                  while [ ! -f /opt/app/data/pnpm-workspace.yaml ]; do
+                    sleep 1; elapsed=$((elapsed + 1))
+                    [ "$elapsed" -ge 120 ] && { echo "ERROR: sync timeout" >&2; exit 1; }
+                  done
+                  exec /opt/app/data/packages/platform-ui/scripts/devspace-startup.sh
+              env:
+                - name: NODE_ENV
+                  value: development
+                - name: VITE_API_BASE_URL
+                  value: ""
```
noa-lucent: [major] `VITE_API_BASE_URL=""` will crash the Vite app at startup.

`config.ts#requireEnv` treats empty strings as missing:

```ts
const val = import.meta.env?.[name];
if (typeof val === 'string' && val.trim()) return val;
// ↑ "".trim() is "" which is falsy → falls through
```

Since this isn't Storybook, `resolveStorybookFallback` returns `null`, and `requireEnv` throws:
```
platform-ui config: required env VITE_API_BASE_URL is missing
```

To fix this without modifying `config.ts`, use a value that `requireEnv` accepts and that still results in same-origin API requests through the Vite proxy. For example:

```yaml
- name: VITE_API_BASE_URL
  value: "/api"
```

`deriveBase(rawApiBase, { stripApi: true })` will strip the trailing `/api`, yielding `http://localhost:3000` (or the pod origin), so `http.get('/api/containers')` resolves to a same-origin path — which the Vite `server.proxy` then forwards to `platform-server:3010`.

packages/platform-ui/DEVSPACE.md
```diff
@@ -0,0 +1,73 @@
...
+2. The pipeline patches the existing `platform-ui` Deployment in place (image,
+   security context, resources, command, env) so no clone is created.
+3. DevSpace attaches to the patched deployment with
+   `start_dev --disable-pod-replace` and only configures sync/ports/logs.
+4. The repo is synced into `/opt/app/data` and the startup script installs
+   dependencies, then launches the Vite dev server on `0.0.0.0:3000`.
+5. The dev pod sets `VITE_API_BASE_URL` to an empty string to force same-origin
+   requests; Vite proxies `/api` and `/socket.io` to
+   `http://platform-server:3010` (or `VITE_PROXY_TARGET`).
+6. On exit, the ArgoCD hook restores auto-sync for `platform-ui`.
```
noa-lucent, line 31: [minor] Update this line to reflect the actual value used after fixing the env var in `devspace.yaml`. If changed to `"/api"`, describe it as:

> The dev pod sets `VITE_API_BASE_URL` to `/api` so the UI makes same-origin requests; Vite proxies `/api` and `/socket.io` to `http://platform-server:3010` (or `VITE_PROXY_TARGET`).

noa-lucent (changes requested):
Clean adaptation of the platform-server DevSpace pattern — the structure, security context, sync config, hooks, and documentation all look good.

One blocking issue: `VITE_API_BASE_URL=""` in `devspace.yaml` will cause a runtime crash because `config.ts#requireEnv` rejects empty strings. See inline comment for the fix.

casey-brooks:
## Test & Lint Summary
- `pnpm --filter @agyn/platform-ui exec tsc -p tsconfig.node.json --noEmit`
- `pnpm --filter @agyn/platform-ui test` (Tests: 85 files, 448 passed, 0 failed, 0 skipped)
- `pnpm --filter @agyn/platform-ui lint` (Lint: no errors)
- `pnpm --filter @agyn/platform-ui run typecheck` (Typecheck: no errors)

packages/platform-ui/devspace.yaml
rowan-stein: Fixed — changed to `/api` in the latest commit.

packages/platform-ui/DEVSPACE.md
rowan-stein: Updated DEVSPACE.md to reflect the `/api` value.

noa-lucent (dismissed):
Both issues verified in the code and resolved:

1. `VITE_API_BASE_URL` correctly changed to `"/api"` — `config.ts#requireEnv` will accept it, and `deriveBase` strips the `/api` suffix to produce a same-origin base URL that works with the Vite proxy.
2. `DEVSPACE.md` updated to match.

Clean, well-structured DevSpace configuration. LGTM.

casey-brooks:
## Update
- Added JSON patch step in `packages/platform-ui/devspace.yaml` to remove liveness/readiness probes before `start_dev`.

## Test & Lint Summary
- `pnpm --filter @agyn/platform-ui exec tsc -p tsconfig.node.json --noEmit`
- `pnpm --filter @agyn/platform-ui test` (Tests: 85 files, 448 passed, 0 failed, 0 skipped)
- `pnpm --filter @agyn/platform-ui lint` (Lint: no errors)
- `pnpm --filter @agyn/platform-ui run typecheck` (Typecheck: no errors)

noa-lucent (dismissed):
Probe removal fix verified. The JSON Patch approach is the correct solution — strategic merge patches can't delete fields, so a separate `--type=json` step with `op: remove` is needed.

The implementation is sound:
- Index `0` is reliable: the Helm chart defines a single main container, init containers live in a separate array, and `extraContainers` (empty by default) would append after index 0.
- The `|| echo "Note: ..."` fallback is appropriate here — it handles the case where the Helm chart is later updated to remove probes, preventing a pipeline failure on a non-critical step.
- Pipeline ordering (strategic merge → JSON patch → start_dev) is correct.

**Minor note (non-blocking):** `platform-server` has the same probe definitions in its Helm values but no removal step in its `devspace.yaml`. If NestJS doesn't serve `/healthz`/`/readyz` in dev mode, it may have the same latent issue. Worth a follow-up outside this PR.

LGTM.

casey-brooks:
## Update
- Added an emptyDir volume + mount for `/opt/app/data` in `packages/platform-ui/devspace.yaml` to ensure UID 1000 can write during sync.

## Test & Lint Summary
- `pnpm --filter @agyn/platform-ui exec tsc -p tsconfig.node.json --noEmit`
- `pnpm --filter @agyn/platform-ui test` (Tests: 85 files, 448 passed, 0 failed, 0 skipped)
- `pnpm --filter @agyn/platform-ui lint` (Lint: no errors)
- `pnpm --filter @agyn/platform-ui run typecheck` (Typecheck: no errors)

noa-lucent (dismissed):
Volume mount fix verified. The approach is correct:

- **Root cause**: CRI creates the `workingDir` (`/opt/app/data`) as root-owned before the container starts. Without a volume, the node user (UID 1000) can't write to it, so DevSpace sync fails with `permission denied`.
- **Fix**: An `emptyDir` volume at `/opt/app/data` combined with `fsGroup: 1000` ensures Kubernetes sets group ownership to GID 1000, making the path writable.
- **Strategic merge safety**: `volumes` merge by `name` and `volumeMounts` merge by `mountPath`, so existing Helm-managed mounts (e.g., nginx configMap) are preserved and the new volume is additive.

LGTM.

casey-brooks:
## Update
- Added `--ignore-scripts` to the DevSpace startup pnpm install to avoid root postinstall/Prisma.

## Test & Lint Summary
- `pnpm --filter @agyn/platform-ui exec tsc -p tsconfig.node.json --noEmit`
- `pnpm --filter @agyn/platform-ui test` (Tests: 85 files, 448 passed, 0 failed, 0 skipped)
- `pnpm --filter @agyn/platform-ui lint` (Lint: no errors)
- `pnpm --filter @agyn/platform-ui run typecheck` (Typecheck: no errors)

noa-lucent (dismissed):
Fix verified. Adding `--ignore-scripts` is the correct solution:

- Root `postinstall` runs `pnpm -r --if-present run prepare`, which triggers platform-server's `prisma generate` — a package not installed in the filtered scope.
- Platform-ui and its workspace dependency (`@agyn/shared`) have no `prepare`/`postinstall`/`preinstall` scripts, so `--ignore-scripts` is safe here.
- Platform-server's devspace doesn't need this flag because it explicitly runs `approve-builds` for prisma before install.

LGTM.

casey-brooks:
## Update
- Raised platform-ui DevSpace container memory requests/limits to 1Gi/4Gi to avoid OOM during install + Vite.

## Test & Lint Summary
- `pnpm --filter @agyn/platform-ui exec tsc -p tsconfig.node.json --noEmit`
- `pnpm --filter @agyn/platform-ui test` (Tests: 85 files, 448 passed, 0 failed, 0 skipped)
- `pnpm --filter @agyn/platform-ui lint` (Lint: no errors)
- `pnpm --filter @agyn/platform-ui run typecheck` (Typecheck: no errors)

packages/platform-ui/DEVSPACE.md
```diff
@@ -0,0 +1,73 @@
...
+- **ArgoCD keeps reverting changes**: Verify the ArgoCD Application exists in
+  the `argocd` namespace and that auto-sync was disabled. Run
+  `kubectl get application platform-ui -n argocd` and restart the session if
+  needed.
+- **Old devspace clone still running**: If `platform-ui-devspace` exists from
+  the previous workflow, delete it with
+  `kubectl delete deployment platform-ui-devspace -n platform`.
+- **OOM kills**: The dev container is limited to 2Gi. Reduce memory usage or
+  increase limits in the cluster if you consistently hit OOMs.
+
```
noa-lucent, line 65: [minor] Stale: this still says "limited to 2Gi" but the limit is now 4Gi.

```md
- **OOM kills**: The dev container is limited to 4Gi. Reduce memory usage or
  increase limits in the cluster if you consistently hit OOMs.
```

noa-lucent (dismissed):
Memory bump verified — request 1Gi / limit 4Gi now matches platform-server's devspace config exactly. Reasonable for pnpm install + Vite dev server.

One minor doc nit: DEVSPACE.md still references the old 2Gi limit in the OOM troubleshooting section. Non-blocking but worth fixing before merge.

casey-brooks:
## Update
- Updated DEVSPACE.md OOM note to reflect 4Gi limit.

## Test & Lint Summary
- `pnpm --filter @agyn/platform-ui exec tsc -p tsconfig.node.json --noEmit`
- `pnpm --filter @agyn/platform-ui test` (Tests: 85 files, 448 passed, 0 failed, 0 skipped)
- `pnpm --filter @agyn/platform-ui lint` (Lint: no errors)
- `pnpm --filter @agyn/platform-ui run typecheck` (Typecheck: no errors)

packages/platform-ui/DEVSPACE.md
rowan-stein: Fixed — updated to 4Gi in latest commit.

casey-brooks:
## Update
- Allowed all hosts for the Vite dev server (`allowedHosts: true`) so `agyn.dev` ingress requests are accepted.

## Test & Lint Summary
- `pnpm --filter @agyn/platform-ui exec tsc -p tsconfig.node.json --noEmit`
- `pnpm --filter @agyn/platform-ui test` (Tests: 85 files, 448 passed, 0 failed, 0 skipped)
- `pnpm --filter @agyn/platform-ui lint` (Lint: no errors)
- `pnpm --filter @agyn/platform-ui run typecheck` (Typecheck: no errors)

## E2E DevSpace Retest
- `git pull origin noa/issue-1389`
- `devspace dev` (platform-ui)
- `curl -H 'Host: agyn.dev' http://localhost:3000` returned HTML:
  ```
  <!doctype html>
  <html lang="en">
  ```

noa-lucent (approved):
Both changes verified:

1. **`allowedHosts: true`** — Correct fix. Vite 6.0.6+ enforces host header validation by default (CVE-2025-30208), and in-cluster requests arrive with the external hostname (`agyn.dev`), which isn't in the default allowed list. Since `server.*` config only applies to `vite dev` / `vite preview` (never production builds) and the dev server runs behind k8s networking, disabling this check is a reasonable trade-off.

2. **DEVSPACE.md** — OOM troubleshooting entry updated from 2Gi to 4Gi. Previous nit resolved.

All prior feedback addressed. LGTM.

vitramir (approved):

